PR #1, opened 2016-02-02
[BROOKLYN-166] Replace existing textareas for editing  YAML with Composer Tab + Codemirror

m4rkmckenna:
Continuatiion of work started in https://github.com/apache/incubator-brooklyn/pull/865 by @azbarcea for [BROOKLYN-166](https://issues.apache.org/jira/browse/BROOKLYN-166)

### DIFF
--- a/src/main/license/source-inclusions.yaml
+++ b/src/main/license/source-inclusions.yaml
@@ -20,10 +20,9 @@
 # extras file for org.heneveld.license-audit-maven-plugin
 # listing projects from which *source* files are included
 
-- id: jquery-core:1.7.2
-- id: swagger-ui:2.1.4
-# we use other versions of the above in other projs
-
+- id: jquery-core
+- id: swagger
+- id: swagger-ui
 - id: jquery.wiggle.min.js
 - id: require.js
 - id: require.js/r.js
@@ -40,3 +39,4 @@
 - id: js-yaml.js
 - id: jquery.form.js
 - id: marked.js
+- id: codemirror.js
--- a/src/main/license/source-inclusions.yaml
+++ b/src/main/license/source-inclusions.yaml
@@ -20,9 +20,10 @@
 # extras file for org.heneveld.license-audit-maven-plugin
 # listing projects from which *source* files are included
 
-- id: jquery-core
-- id: swagger
-- id: swagger-ui
+- id: jquery-core:1.7.2
+- id: swagger-ui:2.1.4
+# we use other versions of the above in other projs
+
 - id: jquery.wiggle.min.js
 - id: require.js
 - id: require.js/r.js
@@ -39,4 +40,8 @@
 - id: js-yaml.js
 - id: jquery.form.js
 - id: marked.js
-- id: codemirror.js
+- id: codemirror.{js,css}
+- id: show-hint.{js,css}
+- id: placeholder.js
+- id: anyword-hint.js
+- id: yaml.js
--- a/src/main/webapp/assets/js/router.js
+++ b/src/main/webapp/assets/js/router.js
@@ -20,12 +20,12 @@ define([
     "brooklyn", "underscore", "jquery", "backbone",
     "model/application", "model/app-tree", "model/location", 
     "model/server-extended-status",
-    "view/home", "view/application-explorer", "view/catalog", "view/script-groovy",
+    "view/home", "view/editor", "view/application-explorer", "view/catalog", "view/script-groovy",
     "text!tpl/help/page.html","text!tpl/labs/page.html", "text!tpl/home/server-caution.html"
 ], function (Brooklyn, _, $, Backbone,
         Application, AppTree, Location, 
         serverStatus,
-        HomeView, ExplorerView, CatalogView, ScriptGroovyView,
+        HomeView, EditorView, ExplorerView, CatalogView, ScriptGroovyView,
         HelpHtml, LabsHtml, ServerCautionHtml) {
 
     var ServerCautionOverlay = Backbone.View.extend({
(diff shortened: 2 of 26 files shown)


src/main/license/source-inclusions.yaml
```diff
@@ -20,10 +20,9 @@
 # extras file for org.heneveld.license-audit-maven-plugin
 # listing projects from which *source* files are included
 
-- id: jquery-core:1.7.2
-- id: swagger-ui:2.1.4
-# we use other versions of the above in other projs
-
```
ahgittin: i think this is an accidental reversion.  the version numbers should be included.  (you might want also to auto-regen the license file -- that also has some wanted new stuff removed but it's not a big deal.)

src/main/webapp/assets/js/router.js
```diff
@@ -135,6 +137,7 @@ define([
         },
 
         showView: function(selector, view) {
+            console.log("showView");
```
ahgittin: remove `log` statement

src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -113,6 +113,7 @@ define([
         if ($input.attr("type") === "checkbox") {
             return $input.is(":checked");
         } else {
+            // TODO: get codemirror.getValue() in case this is a textarea ...
```
ahgittin: can you elaborate?  (or just say something like "comment from @azbarcea, not clear what it means but left in case that becomes clear one day")

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -33,14 +33,17 @@ define([
     "text!tpl/app-add-wizard/deploy-version-option.html",
     "text!tpl/app-add-wizard/deploy-location-row.html",
     "text!tpl/app-add-wizard/deploy-location-option.html",
+    // ↓ not part of the constructor
```
ahgittin: if we're going to start using unicode U+2193 why not go all the way and start using U+1F4A9 as variables in code?

    Injector 💩 = Guice.createInjector(new My̰̰̰̰̰̰̰̰̰̰̰̰̰̰̰̆̆̆̆̆̆̆̆̆̆̆̆̆̆̆̆̆̆ZalgoModule());

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -159,6 +163,7 @@ define([
             this.updateButtonVisibility();
         },
         updateButtonVisibility:function () {
+            log("::updateButtonVisibility()")
```
ahgittin: log statement

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -111,6 +114,7 @@ define([
             this.model.yaml = "";
             this.model.mode = "template";  // or "yaml" or "other"
             this.currentStep = 0;
+            log("ModalWizard::initialize()");
```
ahgittin: and another

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -262,12 +266,16 @@ define([
             this.renderCurrentStep();
         },
         nextStep:function () {
+            log("ModalWizard::nextStep()");
```
ahgittin: "

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -305,15 +313,13 @@ define([
             }
         },
         previewStep:function () {
+            log("ModalWizard::previewStep()");
```
ahgittin: i'm going to stop marking them now.  there are others.

src/main/license/source-inclusions.yaml
```diff
@@ -40,3 +39,4 @@
 - id: js-yaml.js
 - id: jquery.form.js
 - id: marked.js
+- id: codemirror.js
```
ahgittin: for LICENSE to generate correctly the detail for codemirror needs to be added to the extras file `brooklyn-dist`

we should tweak that generation to read from other projects.  although that can wait, when we use bower we'll be rewriting it anyway.

ahgittin:
i've done the dull part, tomorrow i'll do the fun part, trying it out.  thanks @m4rkmckenna @azbarcea!

src/main/webapp/assets/js/util/brooklyn-utils.js
m4rkmckenna: comment from @azbarcea ... no further logic requored as all logic is on composer tab will remove comment

m4rkmckenna:
@ahgittin This PR does not address autocomplete ... thought it better to get the basic functionality in fist

Example of basic auto complete [HERE](https://plnkr.co:443/hF1mSssHv9nj1WKEFnET)

ahgittin:
Nice work @m4rkmckenna @azbarcea .  Merged (and combined all the license items as "codemirror" with a list of files.)

It looks like we now have all the old functionality.  A few things I think still need work:

Important:

* Composer should indicate whether I'm adding to catalog or deploying
* In wizard popup:
  * need a way to get to composer when not selecting a catalog item, ie writing new yaml (common user workflow)
  * "preview" button is not working; it could instead become "Open in Composer"
  * probably no need for the tabs?

Nice to have:

* In composer, blueprint-to-deploy mode could have an option "open as catalog item"
* In catalog, could have "Edit" mode for blueprints that have associated yaml (would be easy and very nice i think!)

Probably not worth it:

* In composer, if in catalog mode, with a single item, we could have an "open as blueprint" mode

ahgittin:
One more important item:
* The dialog should track whether it is a catalog item or a blueprint. Otherwise someone might deploy when they intended to add to the catalog!

ahgittin:
(This will help prevent autodetect mode problems, which happen for instance if you have a comment on a line before a `brooklyn.catalog:` line!)

ahgittin:
Changes are done in #5 and #7

PR #3, opened 2016-02-10
use local version of zeroclipboard movie

ahgittin:
fixes issues in some recent browser (eg latest chrome on latest os x)
where the flash "movie" won't load unless it is local.
the error it gives is a generic "uncaught exception" in
ZeroClipboard activate -> flashState.bridge.setSize.
curiously it still works fine in some browsers (firefox) and
older chrome was also fine but I'm guessing
security modes are being tightened in browsers.

none of the other recommended settings for fixing this seem to work:
* trustedDomains
* trustedOrigin
* allowScriptAccess
* and even the flash config in the browser

it's just a small movie however so should be low impact (smaller than most of our images).
put in img path to simplify access.

also, zeroclipboard is updated from 1.3.1 to 1.3.5 (latest stable 1.x branch).
(2.x branch looks like it might be a bigger api change although some fixes in this area are reported.)


richardcloudsoft:
We cannot include binary files in our releases. ZeroClipboard.swf caused us to fail our incubator releases - see [[1](https://mail-archives.apache.org/mod_mbox/brooklyn-dev/201411.mbox/%3CCABQFKi1qEgVs%2B969Z%3DQ57_C-iPmkNHCYN7kgKEODM8TYQpogaw%40mail.gmail.com%3E)]

If ZeroClipboard does not work from a CDN then we need to either build it from source as part of our build process, or find something else to replace it.

ahgittin:
good spot.  how about we download it from the CDN as part of the build so it isn't part of the source release but it is part of the binary release where it is needed locally?  in that respect it would be the same as downloading JARs so should be fine?  (and that's really what this is.)

closing for now

richardcloudsoft:
Hmm, that's a *very* interesting idea! As you say, comparable to Maven downloading and bundling JAR dependencies - I think it'll fly...

PR #5, opened 2016-02-11
Codemirror and Add Application Wization enhancements

ahgittin:
fixes the issues described at #1 

it's now rather pleasurable to use the yaml editor, and the flow from the wizard to the composer to the catalog and back seems smooth to me

thanks @tbouron and @hellberg for the ui advice on codemirror, it acts almost as a tutor for the different modes (although it would be nice to find a way to make the distinctions even clearer)


m4rkmckenna:
LGTM

PR #7, opened 2016-02-16
composer yaml completion for apps and catalog

ahgittin:
builds on #5 -- so just the last commit

as described in that commit and in comments this is a simple heuristic approach.  it is much more work to keep a parse tree and compare against a schema -- would be nice to do but this gives more bang for the buck, even if it is far from foolproof.


m4rkmckenna:
comments can be addressed later if ever ...

good to merge

ahgittin:
thanks @m4rkmckenna -- good suggestions, i've applied them and merged

PR #8, opened 2016-02-18
use local version of zeroclipboard movie

ahgittin:
fixes issues in some recent browser (eg latest chrome on latest os x)
where the flash "movie" won't load unless it is local.
the error it gives is a generic "uncaught exception" in
ZeroClipboard activate -> flashState.bridge.setSize.
curiously it still works fine in some browsers (firefox) and
older chrome was also fine but I'm guessing
security modes are being tightened in browsers.

none of the other recommended settings for fixing this seem to work:
* trustedDomains
* trustedOrigin
* allowScriptAccess
* and even the flash config in the browser

it's just a small movie however so should be low impact (smaller than most of our images).
put in img path to simplify access.

also, zeroclipboard is updated from 1.3.1 to 1.3.5 (latest stable 1.x branch).
(2.x branch looks like it might be a bigger api change although some fixes in this area are reported.)

this does not include the ZeroClipboard.swf binary but it will download it as is
normal practise for binary dependencies.  slightly unusually this uses maven to download it explicitly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -220,34 +220,49 @@
                 </plugin>
                 <!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
                 <plugin>
-                	<groupId>org.eclipse.m2e</groupId>
-                	<artifactId>lifecycle-mapping</artifactId>
-                	<version>1.0.0</version>
-                	<configuration>
-                		<lifecycleMappingMetadata>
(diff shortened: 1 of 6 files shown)


ahgittin:
this addresses the very valid objection by @rdowner in #3

ahgittin:
merging as this is primarily addressing a previous PR

PR #9, opened 2016-02-24
Fix tests

ygy:
- Rest API fixtures are obtained via maven dependency plugin
- Fixtures' path in no longer relative to the brooklyn-server dir,
  that's why they are obtained as dependency


ahgittin:
@ygy good job figuring out the problem

however `brooklyn-ui` is built before `brooklyn-server` so this isn't a good fix, not for a clean system.  it will work if you have older json files but that's it.

the simplest things to do would be to move the json files to the `ui` project, and have that build bundle them into a test artifact which api/server/etc whichever need them can access.

another option would be to create a `brooklyn-api` top-level apache project.  this could include `api` and some of `rest/rest-api` and would be built before the `ui`.  that's maybe cleaner and it means anything which changes an api class is explicitly an extra pull request.  it would facilitate moving to a swagger-defined api from which our java rest client is generated, and possibly JS clients in future ... but we're a long way from that.  also our `api` project includes utils and test-support which might get updated more often; otoh it might be nice to move them out so that `api` is cleaner.  `rest-api` looks like it includes a lot of helper classes so cleaning that up is a bigger task, hence suggesting splitting it.

none of this adds much value to the user so i'm inclined towards the first suggestion.

@aledsage @hzbarcea @andreaturli @neykov @rdowner thoughts on this?

neykov:
Adding another option - build brooklyn-jsgui after brooklyn-server; don't bundle jsgui with launcher, use it from `libs/wars` instead. With that in mind -1 for a new top-level repo.

Reversing the dependency and having the fixtures in `jsgui` also sounds like a quick win.

ahgittin:
@ygy how's this looking?

ahgittin:
see comments at #51.  in short, merging.

PR #10, opened 2016-02-28
Request container authentication

neykov:
Follow up to https://github.com/apache/brooklyn-server/pull/41

### DIFF
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -1,6 +1,3 @@
-<!DOCTYPE web-app PUBLIC
- "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN"
- "http://java.sun.com/dtd/web-app_2_3.dtd" >
 <!--
     Licensed to the Apache Software Foundation (ASF) under one
     or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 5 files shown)


src/main/webapp/WEB-INF/web.xml
```diff
@@ -19,6 +16,38 @@
     specific language governing permissions and limitations
     under the License.
 -->
-<web-app>
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
```
ahgittin: this file appears duplicated in the rest-server module.  any way to avoid that?

src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -178,7 +178,7 @@ define([
         $.ajax({
             type: "POST",
             dataType: "text",
-            url: "/logout",
```
ahgittin: does the endpoint support `POST` ?  in https://github.com/apache/brooklyn-server/pull/41 it says only `GET`

src/main/webapp/index.html
```diff
@@ -47,7 +47,7 @@
 
 <div class="navbar navbar-fixed-top">
     <div class="navbar-inner">
-        <div class="userName-top"><span id="user"></span> | <a href="/logout" id="logout-link">Log out</a></div>
+        <div class="userName-top"><span id="user"></span> | <a href="/v1/logout" id="logout-link">Log out</a></div>
```
ahgittin: this won't work with the portable ui fix #11 will it?  (this is definitely already an improvement but if there's a further quick fix here or in #11 that would be nice)

ahgittin: maybe define a JS function to ajax get the logout and have this call to that?

src/main/webapp/WEB-INF/web.xml
neykov: There's a `web.xml` in `rest-server`, but it's not a duplicate. This `web.xml` is used during both classic & OSGi deployments. The REST API is overlayed, as a servlet in the same WebContext in classic and as a CXF context plugin in OSGi (both cases used as jar, not war).
The `web.xml` in `rest-server` is never used in the standard brooklyn init. It's there just for `BrooklynRestApiLauncher` which supports both servlet & war deployment modes. For the war case the jsgui is not installed at all.

src/main/webapp/assets/js/util/brooklyn-utils.js
neykov: Good catch, didn't realize there's something more to the logout link at the time.

neykov: Addressed in #41

src/main/webapp/index.html
neykov: It will because the link is just a placeholder, the real work is done by JS, overriding the click (see above comment). The JS is doing a `$.ajax` POST` so it will go through the url patching in #11.

src/main/webapp/assets/css/base.css
```diff
@@ -503,6 +503,10 @@ input[type="color"]:focus,.uneditable-input:focus {
     white-space: nowrap;
 }
 
+.navbar_main error-message {
+    white-space: normal;
```
ahgittin: about to merge -- but i'll remove this, in light of #21

PR #11, opened 2016-02-29
Allow relocation of Brooklyn REST API

CMoH:
This PR follows the proposal on the mailing list about allowing the REST API to be relocated relative to brooklyn's web UI.

For now the function is not changed, but this allows administrators to deploy Brooklyn's REST API and its UI indepentent of each other's base URLs.

### DIFF
--- a/src/main/webapp/assets/js/config.js
+++ b/src/main/webapp/assets/js/config.js
@@ -100,3 +100,17 @@ require([
 ], function (Router) {
     new Router().startBrooklynGui();
 });
+
+/*
+ * Prepend a base URL to REST API calls
+ */
+$.ajaxSetup({
+    beforeSend: function(jqXHR, settings) {
+//        var baseURL = "/api/brooklyn/";
+        var baseURL = "";
+
+        if (baseURL && settings.url.startsWith("/v1")) {
+            settings.url = (baseURL + settings.url).replace("//", "/");
+        }
+    }
+});


src/main/webapp/assets/js/config.js
```diff
@@ -100,3 +100,17 @@ require([
...
+
+/*
+ * Prepend a base URL to REST API calls
+ */
+$.ajaxSetup({
+    beforeSend: function(jqXHR, settings) {
+//        var baseURL = "/api/brooklyn/";
+        var baseURL = "";
+
+        if (baseURL && settings.url.match(/^\/v1\/\w+/i)){
```
neykov: minor, `.indexOf('/v1/') == 0` will be more clear

CMoH:
Actually the credit goes to @sliceratwork - I merely wrote it on my machine.

I don't know about the `/v1` prefix. This is more about allowing various configurations during deployment. I would prefer having an external config for the base URL, but this is a start.

src/main/webapp/assets/js/config.js
sliceratwork: True it would be more clear, but the way it's currently set up is to intercept only calls to URLs that begin with **/v1/**, so something like **/v1/projects** will be affected, while **/api/v1/projects** won't.

If you use `.indexOf('/v1/') == 0`, **/api/v1/projects** will be affected as well.

ahgittin: @sliceratwork i don't think so:

```
% node
> '/api/v1/project'.indexOf('/v1/')
4
```

btw we extend the prototype of `String` in `brooklyn-utils.js` so you could just write  `.startsWith('/v1/')`.

ahgittin:
i like.

in terms of future direction allowing the base url to be configurable at deploy time would be nice, though maybe difficult (?).

minor preference but in future might be cleaner to swap all the `/v1/...` api calls in JS to be `/brooklyn/v1/...` and then we replace any link starting `/brooklyn/` with the new prefix

ahgittin:
PS good to merge once the regex / `startsWith` comments are agreed

CMoH:
I switched to using `startsWith`. Thanks for the tip

PR #13, opened 2016-03-03
Ui fixes

ahgittin:
fixes several minor irritants in the ui

* ui auto-refresh bug: brooklyn-233, dynamic additions are not shown
* ui app-tree reload bug - selected entity hidden if you reload when non-root entity selected
* ui back button bug - tree view hid if you hit back button
* ui activities bug - stale data shown switching tabs and entities and going in to sub-tasks
* composer scrolling bug - the editor would hide the cursor if you do pg-down then pg-up on a large yaml
* composer autocomplete - now completes `location` after `services:` block, and filters unnecessary root proposals

### DIFF
--- a/src/main/webapp/assets/css/codemirror-brooklyn.css
+++ b/src/main/webapp/assets/css/codemirror-brooklyn.css
@@ -19,11 +19,15 @@
 
 /* overrides for brooklyn */
 
+.composer-editor {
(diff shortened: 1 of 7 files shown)


tbouron:
Tested. LGTM.

Just minor type to fix.

PR #17, opened 2016-03-16
Location wizard

tbouron:
Add a location wizard:
wizard

This will be display on:
1. the home page if there is no locations defined. The location number becomes a button to add a location. The wizard will be displayed as a modal in this case
home
2. the catalog page, when clicking on the add location button
catalog
3. the YAML editor page if there is no locations defined. The wizard will be displayed as a modal in this case
editor

### DIFF
--- a/src/main/license/source-inclusions.yaml
+++ b/src/main/license/source-inclusions.yaml
@@ -43,4 +43,5 @@
 - id: codemirror.js
 - id: font-awesome-fonts
 - id: font-awesome-code
+- id: jquery.easy-autocomplete.js
 
--- a/src/main/webapp/assets/css/base.css
+++ b/src/main/webapp/assets/css/base.css
@@ -1490,6 +1490,10 @@ textarea.param-value {
 .location-type-container {
     overflow: hidden;
 }
+.help-text {
+    padding-top: 10px;
+    font-style: italic;
+}
 .location-type {
     width: 160px;
     height: 200px;
--- a/src/main/webapp/assets/css/base.css
+++ b/src/main/webapp/assets/css/base.css
@@ -361,17 +361,16 @@ input[type="color"]:focus,.uneditable-input:focus {
     vertical-align: middle;
 }
 
-.addApplication {
+.addSummary {
     border: 1px solid #a1cb8c !important;
     color: #505050 !important;
     background: url(../images/addApplication-plus.png) no-repeat !important;
     padding: 10px 0px 0px 74px !important;
     width: 298px !important;
     height: 201px !important;
-    margin-right: 0;
 }
 
-.addApplication:hover {
+.addSummary:hover {
     border: 1px solid #58a82e !important;
     color: #58a82e !important;
     background: url(../images/addApplication-plus-hover.png) no-repeat
@@ -1488,6 +1487,9 @@ textarea.param-value {
 }
 
 /* Location wizard */
+.location-type-container {
+    overflow: hidden;
+}
 .location-type {
     width: 160px;
     height: 200px;
@@ -1517,16 +1519,22 @@ textarea.param-value {
 .location-type .location-type-title {
     font-size: 30px;
 }
+input.location-other {
+    margin-top: -10px;
+    margin-left: 10px;
+}
 #catalog-add-form .modal-header .close {
     display: none;
 }
 #catalog-add-form .modal-body {
     max-height: none;
+    overflow: visible;
 }
 #catalog-add-form .modal-body .tab-content {
     min-height: 0;
     background-color: transparent !important;
     padding: 0;
+    overflow: visible;
 }
 #catalog-add-form .location-type {
     border-color: #ddd;
@@ -1540,4 +1548,17 @@ textarea.param-value {
 }
 #catalog-add-form .modal-footer {
     background-color: transparent;
+}
+.easy-autocomplete {
+    display: inline;
+}
+.easy-autocomplete-container ul {
+    margin: 0
+}
+.easy-autocomplete-container ul .eac-category {
+    color: #fff;
+    background-color: rgba(120, 180, 70, 0.8);
+}
+button.remove-entry {
+    margin-top: -10px;
 }
--- a/src/main/webapp/assets/css/base.css
+++ b/src/main/webapp/assets/css/base.css
@@ -1486,3 +1486,34 @@ textarea.param-value {
     color: transparent;
     text-shadow: 0 0 5px rgba(0,0,0,0.5);
 }
+
+/* Location wizard */
+.location-type {
+    width: 160px;
+    height: 200px;
+    float: left;
+    text-align: center;
+    border: 3px solid #EEE;
+    position: relative;
+    margin: 4px;
+}
+.location-type:hover {
+    border-color: #CCC;
+}
+.location-type.selected {
+    border-color: #793;
+}
+.location-type div {
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+    width: 150px;
+    padding: 5px;
+}
+.location-type .fa {
+    font-size: 60px;
+    margin-bottom: 20px;
+}
+.location-type .location-type-title {
+    font-size: 30px;
+}
--- a/src/main/webapp/assets/css/base.css
+++ b/src/main/webapp/assets/css/base.css
@@ -1517,3 +1517,27 @@ textarea.param-value {
 .location-type .location-type-title {
     font-size: 30px;
 }
+#catalog-add-form .modal-header .close {
+    display: none;
+}
+#catalog-add-form .modal-body {
+    max-height: none;
+}
+#catalog-add-form .modal-body .tab-content {
+    min-height: 0;
+    background-color: transparent !important;
+    padding: 0;
+}
+#catalog-add-form .location-type {
+    border-color: #ddd;
+}
+
+#catalog-add-form .location-type:hover {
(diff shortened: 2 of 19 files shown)


tbouron:
Related PR: https://github.com/apache/brooklyn-server/pull/62 => To be merge atfer this one.

ahgittin:
The maven build fails -- same error reported by apache jenkins as on my machine -- due to:

```
[ERROR - 2016-03-17T12:05:21.370Z] Session [85626510-ec38-11e5-968b-4b6a977c9118] - page.onError - msg: Error: Script error
http://requirejs.org/docs/errors.html#scripterror

  phantomjs://platform/console++.js:263 in error
[ERROR - 2016-03-17T12:05:21.370Z] Session [85626510-ec38-11e5-968b-4b6a977c9118] - page.onError - stack:
  onError (http://localhost:53270/src/js/libs/require.js:32)
  A (http://localhost:53270/src/js/libs/require.js:15)
  onScriptError (http://localhost:53270/src/js/libs/require.js:32)

  phantomjs://platform/console++.js:263 in error
```

I'll test this in IDE but the above can be fixed in parallel.

ahgittin:
BTW:  the previous comment does not reflect my huge excitement at this PR!

ahgittin:
We will require to add a reference to `easy-autocomplete` for licensing purposes -- ping me if you want help with that.

ahgittin:
builds fine now, and works well as far as i've been able to test it.  some comments so far...

technical:

* most of the time autocomplete does not show, e.g. on minRam; is there a way to make it more obvious?

cosmetic:

* capitalization is inconsistent.  for titles/labels the pattern has  been "Location Name" (all words in initial caps, acronyms all caps) not "location name" or "Location name"
* more help text is required; imagine a new user -- they won't know what to put for "location id"; could say "Give a label to identify this location. Typically this is lower case using hyphens and no spaces, such as my-openstack-ireland" and similarly for "name" ("Optionally provide a display name for this location, using capitals and spaces, such as 'My OpenStack (Ireland)'."), and "identity" / "credential" it isn't clear this is to log on to the cloud
* how will a user know what BYON is?  (a circle icon giving popup help could work for this and the previous, if too much text is not wanted; alternatively we could call it "Manual IP Addresses")
* the message about "Required information about your location" is cryptic and in fact wrong as many of the fields are not required; remove it altogether?
* the "additional configuration" page is imposing; can we make it clear that there is no need to set anything, but if you want to put constraints or special login there are many config keys available (possibly with link to the website)

conceptual:

* I don't think `localhost` should be a top-level item.  it's only something advanced users should try.  extending an existing location is probably more important.  we could call that tab "Advanced" and within it a text field for `type` with autocomplete allowing `localhost` and any named locations the user has previously set up.

a huge leap forward!

tbouron:
Added the reference to `easy-autocomplete`, see my last commit

tbouron:
Technical:
- What is the issue with the autocomplete? Works ok after at least one character for me.

Cosmetic:
- Will change for all words in initial caps, acronyms all caps
- Will provide helper texts
- Same as above
- Will remove the text
- Buttons are enabled on this screen which makes it obvious that you don't need to do anything else. But I can change the subtitle from `Provisioning information about your location` to `Optional provisioning information about your location`

Conceptual:
- I disagree with that. Those 3 top level location types are taken from the [Brooklyn docs](https://brooklyn.apache.org/v/latest/ops/locations/index.html) to be consistent!

ahgittin:
> What is the issue with the autocomplete? Works ok after at least one character for me.

maybe it's a huge time delay or my machine.  for me it only popped up once and then vanished.  other times i couldn't get anything to show.

is it possible to have ctrl-space also trigger it, and put a hint that this is supported?  that at least makes it consistent with composer and obvious that it is possible.

> > the "additional configuration" page is imposing; can we make it clear that there is no need to set anything, but if you want to put constraints or special login there are many config keys available (possibly with link to the website)
> Buttons are enabled on this screen which makes it obvious that you don't need to do anything else. But I can change the subtitle from Provisioning information about your location to Optional provisioning information about your location

i find button enablement a *subtle* clue not an *obvious* one.  (*disablement* is an obvious indicator that i *cannot* proceed but if i'm taken to a screen i assume i'm supposed to do something there.  the *obvious* button in this screen is *Add configuration* which is intimidating.)

additionally i'd like us to more user-friendly less techie/jargony.  the phrase *provisioning information about your location* whether optional or not isn't clear to me.  also it's too general in many cases (because the previous screen gave a bunch of provisioning information) and wrong in others (specifying a `loginUser` is not provisioning information, it's information to connect to a provisioned machine).

how about text such as: *In many target locations, additional configuration may be supported. Enter any such options here. Autocomplete for some config options is available using Ctrl-Space. For information on available options consult the [LINK]Brooklyn docs. Alternatively you can skip this step.*

> > I don't think localhost should be a top-level item. it's only something advanced users should try. extending an existing location is probably more important. we could call that tab "Advanced" and within it a text field for type with autocomplete allowing localhost and any named locations the user has previously set up.
> I disagree with that. Those 3 top level location types are taken from the Brooklyn docs to be consistent!

the docs have evolved, and 5 years ago pre-vagrant even `localhost` was common.  they should be updated, not this based on that.  the menu options in the docs are better, with "named locations" as the second choice.

localhost is useful for techies and debugging but not a casual user who increasingly we want to target, who is trying to deploy community blueprints for use.  it is something that an advanced user who is comfortable with `ssh localhost` might do max once.  clouds, byon, and named/extending otoh are things people do routinely.

(incidentally i think because `localhost` is available as a resolver you can specify it as a location in a yaml even if it isn't defined as a type.  so i'm not even sure that most users would want to click that ever!)

@shartzel any input?

tbouron:
> maybe it's a huge time delay or my machine. for me it only popped up once and then vanished. other times i couldn't get anything to show.
>
> is it possible to have ctrl-space also trigger it, and put a hint that this is supported? that at least makes it consistent with composer and obvious that it is possible.

ctrl + space to display the autocomplete? It does not work for me. The suggestions are displayed as soon a I type something in the field. If a type a `i`, it display this instantly
screen shot 2016-03-17 at 15 53 15
I think that's an issue with your build.

> i find button enablement a subtle clue not an obvious one. (disablement is an obvious indicator that i cannot proceed but if i'm taken to a screen i assume i'm supposed to do something there. the obvious button in this screen is Add configuration which is intimidating.)
>
> additionally i'd like us to more user-friendly less techie/jargony. the phrase provisioning information about your location whether optional or not isn't clear to me. also it's too general in many cases (because the previous screen gave a bunch of provisioning information) and wrong in others (specifying a loginUser is not provisioning information, it's information to connect to a provisioned machine).
>
> how about text such as: In many target locations, additional configuration may be supported. Enter any such options here. Autocomplete for some config options is available using Ctrl-Space. For information on available options consult the [LINK]Brooklyn docs. Alternatively you can skip this step.

I'll change for your text.

> the docs have evolved, and 5 years ago pre-vagrant even localhost was common. they should be updated, not this based on that. the menu options in the docs are better, with "named locations" as the second choice.
>
> localhost is useful for techies and debugging but not a casual user who increasingly we want to target, who is trying to deploy community blueprints for use. it is something that an advanced user who is comfortable with ssh localhost might do max once. clouds, byon, and named/extending otoh are things people do routinely.
> 
> (incidentally i think because localhost is available as a resolver you can specify it as a location in a yaml even if it isn't defined as a type. so i'm not even sure that most users would want to click that ever!)

I then need to know what to do in details.

ahgittin:
> autocomplete

seems to be an issue with chrome.  works fine on safari.

on chrome it only displays when i navigate away and then back to the text area, oddly.

the idea with ctrl-space was that this is another way to trigger it, in case of the chrome problem, and for when it is blank and you don't even know one letter to type.

(happy for this to be ignored if ctrl-space isn't easy.)

> advanced mode

details to follow

---

**other items**

* the location display name is optional.  if blank brooklyn will use the id.  being blank should not block "next" enablement.  

* also region is optional, jclouds will pick one

* the "next" button does not enable until after i've focussed away from the last text box (we can live with this if not easy to change)

ahgittin:
re **advanced mode**

what i have in mind is totally simple, user defines an id and optionally a name, and there is a single third field which is the "Parent Location" (description: "Enter the location spec which this location should extend. This could be of the form `jclouds:aws-ec2` or `named:another-location-identifier` or `localhost`. For more information see [LINK].")

if possible, have autocomplete which includes all the `named:xxx` where `xxx` is everything in the locations model, and also `localhost`.  (others we can hold off on...)

tbouron:
@ahgittin autocomplete => the screenshot I have taken is one Chrome... Would be better if other people could test it then.

tbouron:
> the "next" button does not enable until after i've focussed away from the last text box (we can live with this if not easy to change)
Could fix it with Angular, don't know how to do it with the current stack. It was already complicated enough to have inter-dependency management between fields.

ahgittin:
@tbouron in that case let's ignore the autocomplete popup and the next enablement issues

can't wait for this all to be angularized!  :)

shartzel:
sorry for the late reply; re: localhost, if I'm following, I believe I agree with what @ahgittin is saying. I understand the logic in remaining consistent with the top level cloud v byon v localhost, but I think it should be compressed away to one flow.

tbouron:
@ahgittin Pushed latest changes. Hopefully it includes everything now.

ahgittin:
fantastic -- merging.  minor tweaks to descriptions etc.

PR #18, opened 2016-03-17
Improve catalog locations info and editability

ahgittin:
takes advantage of improvements in https://github.com/apache/brooklyn-server/pull/70 esp in rest api to make the locations shown in catalog more informative and useful, disabling delete for those which can't be deleted, and even allowing editing for those following the new style of adding to the catalog

(should work nicely in conjunction with #17 !)

note this requires https://github.com/apache/brooklyn-server/pull/70

### DIFF
--- a/src/main/webapp/assets/js/view/catalog.js
+++ b/src/main/webapp/assets/js/view/catalog.js
@@ -111,10 +111,50 @@ define([
         },
 
         composeItem: function(event) {
-            // TODO could make this a catalog item
+            // TODO we should store `catalogYaml` in the catalog so these can be edited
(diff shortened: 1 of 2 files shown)


tbouron:
The code looks good but
- I cannot edit any locations I manually added.
- properties don't show up for locations I manually added.

It that normal @ahgittin ?

tbouron:
Oh wait, do I need to run the server with https://github.com/apache/brooklyn-server/pull/70?

ahgittin:
yes, that's what i meant by "requires" :)

tbouron:
Sorry, I'm still not use to the new repo structure and forget things in the way

PR #20, opened 2016-03-18
Move jQuery code inside a closure managed by require

tbouron:
This is mainly to prevent javascript errors with jQuery not available when running Brooklyn from `BrooklynJavascriptGuiLauncher`


ahgittin:
good fix.  merging.

PR #21, opened 2016-03-22
Show error message on top of the YAML editor

bostko:


src/main/webapp/assets/tpl/editor/page.html
```diff
@@ -47,12 +47,12 @@ <h3 style="margin-top: 6px;">Blueprint Composer</h3>
                     </span>
                 </div>
             </div>
+            <div class="error-message label-message hide" style="display:none;">
+                <div class="label-important">ERROR</div>
+                <span class="error-message-text"></span>
+            </div>
             <div class="navbar_main_wrapper">
                 <div class="navbar_main">
-                    <div class="error-message label-message hide" style="display:none;">
```
tbouron: No need for the `style="display:none;"` as the `hide` class already does it

Also I would remove the `label-message` class as the style of it does not match the current look & feel of the editor. Instead, you can customise the `error-message` class like so:
```css
.error-message {
    background-color: #e2e2e2;
    padding: 15px 10px 15px 10px;
    border-left: 1px solid #d3d3d3;
    border-right: 1px solid #d3d3d3;
}
```

bostko: @tbouron I think the last can be done in a separate PR

tbouron: @bostko It should be part of this PR in my opinion as this is a specific style, not a global one. You just need to add the code I mentioned to the `base.css` file :)

PR #23, opened 2016-04-06
BROOKLYN-244: Fix blueprint composer loading issue

tbouron:
This loads straight away CodeMirror when launching the composer from the composer tab, removing the requirement on loading the full list of applications contained within the catalog.

When coming from the catalog (editing a catalog item YAML), it will temporarily disable the editor, add an overlay displaying `Loading item...` until the item is full loaded from the REST server and added to the editor. At that point, the overlay will fade out and the editor enable again.

### DIFF
--- a/src/main/webapp/assets/css/codemirror-brooklyn.css
+++ b/src/main/webapp/assets/css/codemirror-brooklyn.css
@@ -54,11 +54,32 @@ li.CodeMirror-hint-active {
   /* override show-hint to have roughly brooklyn colors instead of blue */
   background: #549e2b;
 }
-#composer .error-message {
+
+.composer .error-message {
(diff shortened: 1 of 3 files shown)


m4rkmckenna:
LGTM

Tested with a simulated 50 kb/s connection ... worked as expected

aledsage:
Thanks @tbouron and @m4rkmckenna 

Merging now. Sounds like something to get into the next 0.9.0 release candidate.

tbouron:
Great, thank @aledsage!

If it gets into 0.9.0, you might also want #24

m4rkmckenna:
@aledsage +1 to including in the 0.9.0 release

PR #24, opened 2016-04-06
Update editor placeholder

tbouron:
The text now makes it clear that users can use drag & drop to add text files' content into the editor

brooklyn js rest client

### DIFF
--- a/src/main/webapp/assets/js/view/editor.js
+++ b/src/main/webapp/assets/js/view/editor.js
@@ -204,7 +204,8 @@ define([
                     hintOptions: {
                         completeSingle: false,
 //                        closeOnUnfocus: false,   // handy for debugging
-                    }
+                    },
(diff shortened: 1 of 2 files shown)


m4rkmckenna:
LGTM

**MINOR**
* Might be worth adding styling for dragover
* Might be worth verifying the type of the dropped file

tbouron:
The drag & drop feature is provided by CodeMirror. I'm not sure it can be customised, I'll need to check that.

That being said, the content of the file is inserted where your cursor / mouse cursor drop the file, so having a classic overlay "drop file here" won't work here.

m4rkmckenna:
My thinking was binding to the dragenter / dragover / dragleave events from the textarea assuming the codemirror code propagates said events.

Then adding / removing a class from the textarea OR changing the drag icon

tbouron:
We will probably have to cheat as the textarea is hidden by CodeMirror. AFAIK, we can still bound to those events but from the CodeMirror container rather than the textarea.

I'll investigate more tomorrow

PR #25, opened 2016-04-07
Hotfix: Skip the creation of the contains method within Array prototype

tbouron:
This avoids breaking other parts of the UI

### DIFF
--- a/src/main/webapp/assets/js/view/location-wizard.js
+++ b/src/main/webapp/assets/js/view/location-wizard.js
@@ -541,7 +541,8 @@ define([
                                     return item.getIdentifierName();
                                 }),
                                 spec: ['localhost']
-                            }
+                            },
+                            loggerEnabled: false
                         });
                     }
                 });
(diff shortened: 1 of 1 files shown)


neykov:
Was just about to reply to the email thread :).
The code causing the problem is called only when `loggerEnabled` is true. Suggest disabling logging instead.
Also can't believe they pollyfilled `contains` just to use it a single time on the line following it.

tbouron:
> The code causing the problem is called only when loggerEnabled is true. Suggest disabling logging instead.

Even though, not removing the `contains` is really a big issue. I feel more confortable removing it as it affect every JS arrays.
Again, this is just a temporary fix, I'll look for another autocomplete library after 0.9.0

> Also can't believe they pollyfilled contains just to use it a single time on the line following it.

And yet, it is sadly true... But it's half of my mistake, I choose this library :(

neykov:
My point is that with `loggerEnabled=false` the `Array` prototype will not be patched with `contains`, so no need to delete it.

neykov:
Also I am sure there are bugs in any library we use (as a general principle) :). The question is how long it takes before bumping into them.

tbouron:
> My point is that with loggerEnabled=false the Array prototype will not be patched with contains, so no need to delete it.

My bad, I get what you are saying now. I'll update the PR

tbouron:
@neykov Done

neykov:
The fix works for me, merging.
Thanks @tbouron.

PR #26, opened 2016-05-19
Location wizard generates correct spec for Openstack

sjcorbett:
The wizard previously used `jclouds:openstack` and adding to the catalogue threw an error with "Transformer for Brooklyn OASIS CAMP interpreter gave an error creating this plan: No class or resolver found for location type jclouds:openstack"


mikezaccardo:
Tested, works as expected.

LGTM

sjcorbett:
@mikezaccardo Thanks. Will merge.

PR #27, opened 2016-05-24
Show config values when jumping from App Wizard to YAML Preview

bostko:
I hope the title explains what it is fixing
screenshot from 2016-05-24 14-57-36

I see that the `StepDeploy.validate` does more than validating the model.
Any suggestions for separating this logic or renaming the validate method to `fillConfigAndValidate`?
I do not understand how `ModalWizard`, `ModalWizard.StepCreate` and `ModalWizard.StepDeploy` are connected to each other.
@tbouron any comments?

### DIFF
--- a/src/main/webapp/assets/js/view/application-add-wizard.js
+++ b/src/main/webapp/assets/js/view/application-add-wizard.js
@@ -303,6 +303,10 @@ define([
         },
         previewStep:function () {
             // no need for validation if going to composer
+
+            // When user jump to the YAML editor he also must have the config values that he set through the UI.
+            // The config is set implicitly when user clicks deploy and StepDeploy.validate is called.
+            this.currentView.model.spec.set("config", this.currentView.getConfigMap());
             if (this.currentStep==0) {
                 // from first step, composer should load yaml from the catalog item
                 if (this.currentView.selectedTemplate && this.currentView.selectedTemplate.id) {


src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -303,6 +303,7 @@ define([
         },
         previewStep:function () {
             // no need for validation if going to composer
+            this.currentView.validate();
```
tbouron: The comment right above is valid, at that point, we shouldn't validate the model but send exactly what the user specified. So rather than calling validate here, I would add this above line `317`:
```js
this.model.spec.set("config", this.currentStep.getConfigMap());
```

tbouron:
Indeed, the config is skipped when we send the blueprint to the YAML editor, I commented on that part.

Regarding the way the steps are connected, you can look at the [global view definition](https://github.com/bostko/brooklyn-ui/blob/b6acf26fd43138e47e37ac816435b971ce5af71a/src/main/webapp/assets/js/view/application-add-wizard.js#L133-L146) and the `nextStep` and `prevStep` functions. Those are responsible to render the right step (backbone views)

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -303,6 +303,7 @@ define([
         },
         previewStep:function () {
             // no need for validation if going to composer
+            this.currentView.model.spec.set("config", this.currentView.getConfigMap());
```
tbouron: Could you just add a comment stating what this does please? Will be useful for the future

tbouron:
Tested, LGTM.

PR #29, opened 2016-06-07
Use password input for sensitive information

bostko:
This PR depends on https://github.com/apache/brooklyn-server/pull/185

### DIFF
--- a/src/main/webapp/assets/tpl/apps/param.html
+++ b/src/main/webapp/assets/tpl/apps/param.html
@@ -34,6 +34,10 @@ under the License.
         <input type="checkbox" class="param-value">
         <% } %>
     </td>
+    <% } else if (shouldSanitize == true) { %>
+    <td>
+        <input type="password" class="input-medium param-value" value="<%- defaultValue %>">
+    </td>
     <% } else { %>
     <td><!--  use 1 line textarea so it can be resized -->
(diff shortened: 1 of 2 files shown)


src/main/webapp/assets/tpl/apps/param.html
```diff
@@ -34,6 +34,10 @@
         <input type="checkbox" class="param-value">
         <% } %>
     </td>
+    <% } else if (sensitive == true && type == 'java.lang.String') { %>
+    <td>
+        <input type="password" class="param-value">
```
sjcorbett: What about the default value? And maybe worth using the `input-medium` class for consistency with the case below.

src/main/webapp/assets/tpl/apps/param.html
```diff
@@ -34,6 +34,10 @@
         <input type="checkbox" class="param-value">
         <% } %>
     </td>
+    <% } else if (sensitive == true && type == 'java.lang.String') { %>
```
aledsage: Do we really need to check for type == String, or should we mask out anything that seems to be sensitive?
As a silly example, there code be a 4-digit pin code (which someone has represented as a number).

Presumably we can just mark it as password, get the value, and then rely on our type coercions like we do everywhere else?

aledsage:
LGTM; merging.

PR #30, opened 2016-08-03
BROOKLYN-323: Improve logout script

bostko:
- fix logout for Internet Explorer
- use get call to logout

### DIFF
--- a/src/main/webapp/assets/js/util/brooklyn-utils.js
+++ b/src/main/webapp/assets/js/util/brooklyn-utils.js
@@ -175,17 +175,16 @@ define([
     };
 
     Util.logout = function logout() {
-        $.ajax({
-            type: "POST",
-            dataType: "text",
-            url: "/v1/logout",
-            success: function() {
-                window.location.replace("/");
-            },
-            failure: function() {
-                window.location.replace("/");
-            }
-        });
+        var ua = window.navigator.userAgent;
+        if (ua.indexOf("MSIE ") >= 0 || ua.indexOf(" Edge/") >= 0 || ua.indexOf(" Trident/") >= 0) {
+            document.execCommand('ClearAuthenticationCache', 'false');
+            window.location.replace('/');
+        } else {
+            $('<form action="' + '/logout_redirect.html' + '" method="POST" id="redirectForm">' +
+                '<input type="hidden" name="acme" value="acme"/>' +
+                '</form>').appendTo($(document.body))
+                .submit();
+        }
     }
 
     Util.setSelectionRange = function (input, selectionStart, selectionEnd) {
@@ -216,7 +215,7 @@ define([
 
     $("#logout-link").on("click", function (e) {
         e.preventDefault();
-        Util.logout()
+        Util.logout();
         return false;
     });
 
--- a/src/main/webapp/logout_redirect.html
+++ b/src/main/webapp/logout_redirect.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Title</title>
+</head>
+<body>
+<script>
+    var a=new window.XMLHttpRequest;
+    a.open('POST','/logout',0,'user','wrong'+(new Date).getTime().toString());a.send('');
+    window.location.href='/';
+</script>
+</body>
+</html>


nakomis:
Tested alongside https://github.com/apache/brooklyn-server/pull/288 and this resolves the issue on Windows 10 on both IE and Edge. Behaviour on Chrome on OS X remains unchanged

LGTM

src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -175,17 +175,22 @@ define([
...
-        });
+        var ua = window.navigator.userAgent;
+        if (ua.indexOf("MSIE ") > 0 || ua.indexOf(" Edge/") > 0 || ua.indexOf(" Trident/") > 0) {
+            $.ajax({
+                async: false,
+                type: "GET",
+                dataType: "text",
+                url: "/v1/logout"
+            });
+            document.execCommand('ClearAuthenticationCache', 'false');
```
neykov: Instead of doing a user agent check can we call `ClearAuthenticationCache` in a try-catch and use the code below for all browsers?

bostko: The command ClearAuthenticationCache it's compatible only for IE. What's the point in trying it for other browsers?

src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -175,17 +175,22 @@ define([
...
-                window.location.replace("/");
-            },
-            failure: function() {
-                window.location.replace("/");
-            }
-        });
+        var ua = window.navigator.userAgent;
+        if (ua.indexOf("MSIE ") > 0 || ua.indexOf(" Edge/") > 0 || ua.indexOf(" Trident/") > 0) {
+            $.ajax({
+                async: false,
```
neykov: I'd avoid this at all costs. If the server is stalled or unreachable for a short while the whole UI will freeze. See other comment for doing away with this call altogether.

src/main/webapp/assets/js/util/brooklyn-utils.js
neykov: To avoid the (potentially brittle) user agent check - it will fail but it doesn't matter when in a try-catch block. And when it works it will do its thing.

src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -175,17 +175,23 @@ define([
...
-                window.location.replace("/");
-            },
-            failure: function() {
-                window.location.replace("/");
-            }
-        });
+        var ua = window.navigator.userAgent;
+        if (ua.indexOf("MSIE ") > 0 || ua.indexOf(" Edge/") > 0 || ua.indexOf(" Trident/") > 0) {
+            $.ajax({
+                type: "GET",
```
neykov: `logout` is a `POST` operation.

aledsage:
jenkins test failure is (presumably) unrelated: `java.io.IOException: No space left on device`.

aledsage:
@m4rkmckenna @sjcorbett can you please review and merge?

neykov:
Looks good. To add some background on the unusual logout implementation - the server-side implementation will work correctly only if changing user names after logout. If the user tries to login with the same username it will keep prompting him for password. More details in the server code.
This is papering over the problem until we implement form based auth.

src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -175,17 +175,16 @@ define([
...
-            url: "/v1/logout",
-            success: function() {
-                window.location.replace("/");
-            },
-            failure: function() {
-                window.location.replace("/");
-            }
-        });
+        var ua = window.navigator.userAgent;
+        if (ua.indexOf("MSIE ") > 0 || ua.indexOf(" Edge/") > 0 || ua.indexOf(" Trident/") > 0) {
```
m4rkmckenna: Should this be `>=`?

bostko: Theoretically yes but in reality it is always like `* \(Windows ... (MSIE|Edge)`.
However changing it to `>=`.

m4rkmckenna:
LGTM

PR #31, opened 2016-08-12
show tabs in CM editor

ahgittin:
based on https://codemirror.net/demo/visibletabs.html


ahgittin:
This just makes them visible as in the link above.  We should give a better error message and go to the location but this super-easy fix will help.

m4rkmckenna:
LGTM / Tested

shartzel:
good idea

PR #34, opened 2016-09-28
Use new pinned and constraints flag returned by the REST API

tbouron:
This allows the UI to work out which parameters need to be displayed on the app wizard, following this convention:
- if a parameter if marked as `pinned`, it is always displayed
- if a parameter has a `required` constraint and empty `defaultValue`, it is displayed
Everything is sorted by the priority


src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -680,8 +680,13 @@ define([
                 this.$('.required-config-loading').hide()
                 if (catalogEntryItem!=null && catalogEntryItem.config!=null) {
                     var that = this
-                    _.each(catalogEntryItem.config, function (cfg) {
-                        if (cfg.priority !== undefined) {
+                    _.chain(catalogEntryItem.config).sortBy(function (cfg){
+                        return cfg.priority ? cfg.priority : 99999;
```
neykov: `priority = 0` is a valid value

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -680,8 +680,13 @@ define([
...
+                    }).each(function (cfg) {
+                        if (cfg.pinned === true) {
+                            var html = _.template(RequiredConfigEntryHtml, {data:cfg});
+                            that.$('.config-table').append(html);
+                        } else if (cfg.constraints.indexOf('required') > - 1 && (!cfg.defaultValue || cfg.defaultValue === '')) {
```
neykov: Both cases have the same body so can merge the check by `||`-ing the two alternatives (two make it clear that teartment is the same)
`''` is treated as false so could change `(!cfg.defaultValue || cfg.defaultValue === '')` to `!cfg.defaultValue`.

neykov:
Needs fixing the `priority=0` case, the rest is nitpicks.

src/main/webapp/assets/js/view/application-add-wizard.js
tbouron: > '' is treated as false

Are you sure of that? I always check if a variable exists **and** different than empty because I had troubles in the past. I would update for:
```js
if (cfg.pinned === true || (typeof cfg.defaultValue === 'string' && cfg.defaultValue !== '')) {
    var html = _.template(RequiredConfigEntryHtml, {data:cfg});
    that.$('.config-table').append(html);
}
```

tbouron:
@neykov Comments addressed

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -680,10 +680,12 @@ define([
                 this.$('.required-config-loading').hide()
                 if (catalogEntryItem!=null && catalogEntryItem.config!=null) {
                     var that = this
-                    _.each(catalogEntryItem.config, function (cfg) {
-                        if (cfg.priority !== undefined) {
+                    _.chain(catalogEntryItem.config).sortBy(function (cfg){
+                        return !isNaN(cfg.priority) ? cfg.priority : 99999;
```
neykov: Nitpick, The REST API is configured with `@JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)` so `null` values server-side won't be included in the response (will be `undefined`). Given that the original check `cfg.priority !== undefined` is more explicit - for example `isNaN(null) == false` so not really clear what's the purpose of the check when reading the code.

src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -680,10 +680,12 @@ define([
...
+                    }).each(function (cfg) {
+                        if (cfg.pinned === true || (cfg.contraints.indexOf('required') > -1 && typeof cfg.defaultValue === 'string' && cfg.defaultValue === '')) {
```
neykov: This doesn't handle `defaultValue == null` (or is null converted to `''` somethere?).
Could do:
```
cfg.pinned || (cfg.contraints.indexOf('required') > -1 && (cfg.defaultValue === undefined || cfg.defaultValue === ''))
```

`undefined` because `null` values won't be included in the response.

Would be weird to have a `defaultValue` the empty string and required at the same time, but you never know :).

tbouron: > This doesn't handle defaultValue == null (or is null converted to '' somethere?).
Could do

This would never happens as the REST API returns the `Predicate.toString()`

tbouron: > Would be weird to have a defaultValue the empty string and required at the same time, but you never know :)

That's the point of this check, what if as a blueprint author I want to ensure that one of my parameter has be filled without giving it a default?

src/main/webapp/assets/js/view/application-add-wizard.js
tbouron: Do you really want me to change it again?

src/main/webapp/assets/js/view/application-add-wizard.js
neykov: That's the `defaultValue`, not `constraints`. It can (and will frequently) be null.

tbouron: No it cannot. `contraints` [is always there](https://github.com/apache/brooklyn-server/blob/master/rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java#L69) and empty if there isn't any

neykov: Correct.
`defaultValue` on the other hand can be `null` and is serialized as `undefined`. The check above doesn't handle the `defaultValue is null` case.

tbouron: Sorry but it's not possible. In case of `defaultValue == null` in java, [it is not serialised](https://github.com/apache/brooklyn-server/blob/master/rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/ConfigSummary.java#L45-L46) therefore, the only possibilities are either `undefined` or a non empty string.

neykov: Agree, that's my point.
`defaultValue == null` on the server side is serialized as missing (`undefined`) on the wire.
Which means that `cfg.contraints.indexOf('required') > -1 && typeof cfg.defaultValue === 'string' && cfg.defaultValue === ''` will be `false` for `undefined`. So it will not be shown, even if required.

neykov:
Thanks for the updates @tbouron, merging.

PR #36, opened 2016-11-09
Updated logout step to work in karaf mode

bostko:
- BROOKLYN-323: Simplify LogoutApi
- relative urls for swagger html

PR depends on https://github.com/apache/brooklyn-server/pull/578

Previously when you are on the API page dropdown menu was not visible.
screenshot from 2016-11-09 13-17-35

### DIFF
--- a/src/main/webapp/assets/html/swagger-ui.html
+++ b/src/main/webapp/assets/html/swagger-ui.html
@@ -40,6 +40,8 @@
     <script src='../swagger-ui/lib/backbone-min.js' type='text/javascript'></script>
     <script src='../swagger-ui/lib/swagger-ui.min.js' type='text/javascript'></script>
     <script src='../swagger-ui/lib/marked.js' type='text/javascript'></script>
+    <script src='../js/libs/require.js' type='text/javascript'></script>
+    <script src='../js/libs/bootstrap.js' type='text/javascript'></script>
 
     <script type="text/javascript">
         $(function () {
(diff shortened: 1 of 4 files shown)


tbouron:
Tested, LGTM

neykov:
I think that's the wrong way to fix the problem. The swagger page is opened with `target=_blank` so it's a separate on from the main application. This makes it surprising - say you click on the rest api page which opens a new tab (you could even miss that), then click back on applications. Now you have to identical tabs.
We should either:
  * remove the frame from the swagger page so it's obvious that a new tab is opened. With no way to click on the application tab (which brings even more confusion)
  * Open the swagger ui in the same window

bostko:
@neykov I changed it to open in the same tab.
Can you check again whether it looks consistent UX.

neykov:
LGTM.  This has been changed back and forth in the past, so @ahgittin, @m4rkmckenna any comments?

src/main/webapp/assets/js/config-swagger.js
```diff
@@ -0,0 +1,12 @@
+require.config({
```
ahgittin: where is this file used?

does it need the RAT header?

src/main/webapp/assets/html/swagger-ui.html
```diff
@@ -81,15 +83,15 @@
             <a class="logo" href="#" title="Brooklyn, Version 0.10.0-SNAPSHOT"><!-- Logo added via CSS --></a> <!-- BROOKLYN_VERSION -->
```
ahgittin: the line 2 above, don't we need to change `href="/logout"` to `href="../../v1/logout"` ?

src/main/webapp/assets/js/config-swagger.js
neykov: Somehow related, https://github.com/apache/brooklyn-ui/pull/38 activates the rat plugin for the project. So good point about adding license here.

sjcorbett:
@bostko could you revisit Alex and Svet's comments?

sjcorbett:
@bostko looks good. Will test and merge if ok.

sjcorbett:
@bostko Looks like this negatively impacts logging out of the non-Karaf distribution. After logging out I've found that:
* master non-karaf: prompts for re-authentication.
* master karaf: 405 error `POST is not supported by this URL`
* this pr karaf: displays page with "login" in big text. Clicking it prompts for re-authentication.
* this pr non-karaf: displays page with "login" in big text. Clicking it logs the user back in without requiring re-authentication.

We should prompt the user to re-authenticate in the standard distribution.

bostko:
@sjcorbett thanks for the notes!
I added in apache/brooklyn-server#578 an `unauthorize` call which seems to solve the issues in classic mode.
Could you check both PRs again,
Thanks!

sjcorbett:
@bostko thanks for the update. I've confirmed that the browser requires re-authentication in both cases.

PR #37, opened 2016-11-13
request and set the csrf header protection added to brooklyn server

ahgittin:
Minor tweaks to opt-in to CSRF protection in Brooklyn UI

NB requires https://github.com/apache/brooklyn-server/pull/430
(actually it will work fine without that, but it will be no-op)

### DIFF
--- a/src/main/webapp/assets/js/model/server-extended-status.js
+++ b/src/main/webapp/assets/js/model/server-extended-status.js
@@ -22,6 +22,13 @@ define(["backbone", "brooklyn", "view/viewutils"], function (Backbone, Brooklyn,
         callbacks: [],
         loaded: false,
         url: "/v1/server/up/extended",
+        sync: function(method, collection, options){
+            options = options || {};
+            options.beforeSend = function (xhr) {
+                xhr.setRequestHeader('X-Csrf-Token-Required-For-Requests', 'write');
+            };
+            return Backbone.Model.prototype.sync.apply(this, arguments);
+        },
         onError: function(thiz,xhr,modelish) {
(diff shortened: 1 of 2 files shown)


geomacy:
+1 looks good to me to the best of my knowledge.  See comment on apache/brooklyn-server#430.

bostko:
@ahgittin this breaks logout button :(

ahgittin:
@bostko oh no - pretty sure i tried that but will check

geomacy:
I didn't check it unfortunately - tried other things but don't normally hit the logout button.

bostko:
Sorry for disturbing you I think I am wrong that this exactly broke it.
Please accept that as a reminder to test logout.
Logout htaccess can be broken in numerous ways...

ahgittin:
@bostko what is the problem you're encountering?

i've just tested, running `BrooklynJavascriptGuiLauncher` with this in my `brooklyn.properties`:

```
brooklyn.webconsole.security.users=admin
brooklyn.webconsole.security.user.admin.password=nimda
brooklyn.webconsole.security.provider=org.apache.brooklyn.rest.security.provider.ExplicitUsersSecurityProvider
```

logout seems to work fine in every permutation i've tried.  note there were some changes to `LogoutResource` in https://github.com/apache/brooklyn-server/pull/430 -- i did my testing with them ... they're now in master so shouldn't the issue but might be?

ahgittin:
i see, the problem is the swagger page, @bostko you're addressing in https://github.com/apache/brooklyn-ui/pull/36 -- commented there

(in general logout is working okay and this PR we think is fine)

PR #50, opened 2017-11-20
fix username in quotes

ahgittin:
introduced by #754 returning correct json when advertising json


asfgit:
SUCCESS 
 89 tests run, 0 skipped, 0 failed.
--none--

PR #51, opened 2018-01-03
if not using persistence and mgmt plane ID is null, UI can give erroneous error

ahgittin:
fix it by accepting null plane as acceptable if server is otherwise healthy


geomacy:
looks ok to me, will merge it

PR #55, opened 2018-07-27
Add .m2 folder in the list of ignore folders for the RAT plugin

tbouron:
As `brooklyn-ui` doesn't use `brooklyn-parent`, we need to add the local maven cache for the Jenkins build


ahgittin:
LGTM

PR #58, opened 2018-08-03
Version bumps to align with brooklyn server

ahgittin:
Otherwise it pulls in old version of jackson


duncangrant:
LGTM - I'll merge

PR #62, opened 2018-08-17
Restore slf4j.version property

kemitix:
The brooklyn-ui-utils module wouldn't build on a mvn clean install with this property missing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,6 +105,8 @@
         <node.version>v8.4.0</node.version>
         <npm.version>5.4.1</npm.version>
 
+        <slf4j.version>1.7.14</slf4j.version>
+
     </properties>
 
     <build>


ahgittin:
@kemitix We shouldn't have to redeclare it as it can cause different versions of JARs to be pulled in if they get out of sync (which they have here!).  It is declared in `brooklyn-server/pom.xml` as `1.7.25` (not `.14` used here), which is the parent of `brooklyn-server/parent/pom.xml` which is the parent of this.

The right fix would be to ensure the ancestor definition is pulled in correctly, which it should be.  I am curious why it isn't.  I don't see why this would fix it, as the grandparent pom already declares the variable you declare.  Suggest re-trying with a clean build on `brooklyn-server` then this (without this fix) and seeing whether it is okay.

ahgittin:
@kemitix can you check ^ and update or close this pls?

kemitix:
@ahgittin Retrying build without this fix to confirm if it is needed.

kemitix:
Successfully build from the main `brooklyn` project without this fix in place.

PR #63, opened 2018-08-20
Fix bug with the DSL editor/viewer

tbouron:
This fixes 2 bugs:
1. When adding a DSL for an entity spec, The DSL editor "Done" button was not redirecting to the spec editor. This was due to the previous param object being modified later on by the `ui-router`. Creating a deep copy of the object avoid this
2. The DSL viewer was throwing an error while getting the name of the entity for DSL like `$brooklyn:self()` or `$brooklyn:parent()`. This is fixed by looking up directly for the relationships if the DSL are of that form

### DIFF
--- a/ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
+++ b/ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
@@ -51,7 +51,16 @@ export function dslViewerDirective() {
             return [KIND.STRING, KIND.NUMBER].includes(dsl.kind);
         };
         scope.getRelatedEntity = () => {
-            return scope.dsl.getRoot().relationships.find(entity => entity.id === scope.dsl.params[0].name);
+            if (scope.dsl.params.length > 0) {
+                // If the DSL is looking at an entity ID
+                return scope.dsl.getRoot().relationships.find(entity => entity.id === scope.dsl.params[0].name)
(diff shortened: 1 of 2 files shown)


ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
```diff
@@ -51,7 +51,16 @@ export function dslViewerDirective() {
             return [KIND.STRING, KIND.NUMBER].includes(dsl.kind);
         };
         scope.getRelatedEntity = () => {
-            return scope.dsl.getRoot().relationships.find(entity => entity.id === scope.dsl.params[0].name);
+            if (scope.dsl.params.length > 0) {
+                // If the DSL is looking at an entity ID
+                return scope.dsl.getRoot().relationships.find(entity => scope.dsl.params.length > 0 && entity.id === scope.dsl.params[0].name)
```
ahgittin: no need to check `scope.dsl.params.length > 0` again in 56 after doing it in 54 is there?

ahgittin:
good fixes.  minor comment, good to merge with or without it.

ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
tbouron: Ah good point, forgot to remove this, I'll do right now

PR #64, opened 2018-08-21
Fix configuration for npm plugin on Windows

grkvlt:
Sets the goals explicitly in profile configuration, to match plugin configurations.


ahgittin:
LGTM - thx

PR #66, opened 2018-08-28
recognize memberSpec anywhere, and refresh blueprint within yaml editor

ahgittin:
makes it easier for other entities to work with memberSpecs, in composer or yaml editor

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -79,6 +79,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
         }
         setBlueprintFromJson(newBlueprint);
         $log.debug(TAG + 'Blueprint set from YAML', blueprint);
+        // TODO refresh the blueprint now?  see comments in yaml.state.js and on refreshBlueprint
     }
 
     function getBlueprint() {
@@ -195,6 +196,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
     }
 
     function refreshBlueprintMetadata(entity = blueprint, family = 'ENTITY') {
+        // TODO ideally we'd have a cache for all types
         return refreshEntityMetadata(entity, family).then(()=> {
             return $q.all(entity.children.reduce((result, child) => {
                 result.push(refreshBlueprintMetadata(child));
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -302,13 +302,11 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
 
     function refreshConfigMemberspecsMetadata(entity) {
         let promiseArray = [];
-        if (entity.isCluster()) {
-            Object.values(entity.getClusterMemberspecEntities()).forEach((memberSpec)=> {
-                // memberSpec can be `undefined` if the member spec is not a `$brooklyn:entitySpec`, e.g. it is `$brooklyn:config("spec")`.
-                // there may be a better way but this seems to handle it.
-                if (memberSpec) promiseArray.push(refreshBlueprintMetadata(memberSpec, 'SPEC'));
-            });
-        }
+        Object.values(entity.getClusterMemberspecEntities()).forEach((memberSpec)=> {
+            // memberSpec can be `undefined` if the member spec is not a `$brooklyn:entitySpec`, e.g. it is `$brooklyn:config("spec")`.
+            // there may be a better way but this seems to handle it.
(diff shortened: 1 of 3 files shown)


ui-modules/blueprint-composer/app/views/main/yaml/yaml.state.js
```diff
@@ -46,6 +46,7 @@ function yamlStateController($scope, $rootScope, $timeout, blueprintService, brS
 
             try {
                 blueprintService.setFromYaml(cm.getValue(), true);
+                blueprintService.refreshBlueprintMetadata();
```
tbouron: There is no point of refreshing the metadata here as the YAML state doesn't use this info. When you flip to the graphical view, this method will be called

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
```diff
@@ -115,6 +116,8 @@ export class Entity {
      *
      */
     touch() {
+        // include a summary to aid with debugging (otherwise log just shows the property lastUpdated)
+        this.summary = (this.type || "unset") + (this.id ? " "+this.id : "");
```
tbouron: I don't get the idea behind the summary, we already the type and id within the object. Can you explain this please @ahgittin ?

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
```diff
@@ -19,9 +19,10 @@
 import {Issue} from './issue.model';
 import {Dsl, DslParser} from './dsl.model';
 
-const MEMBERSPEC_REGEX = /^(\w+\.)+[mM]ember[sS]pec$/;
-const FIRST_MEMBERSPEC_REGEX = /^(\w+\.)+first[mM]ember[sS]pec$/;
-const ANY_MEMBERSPEC_REGEX = /^(\w+\.)+(first)?[mM]ember[sS]pec$/;
+const MEMBERSPEC_REGEX = /^(\w+\.)*[mM]ember[sS]pec$/;
+const FIRST_MEMBERSPEC_REGEX = /^(\w+\.)*first[mM]ember[sS]pec$/;
+// TODO ideally we'd just look at type EntitySpec, not key name, but for now look at keyname, anything ending memberSpec
```
tbouron: We cannot as all we have is a YAML file, hence why we look for the key name

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -302,13 +302,11 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
 
     function refreshConfigMemberspecsMetadata(entity) {
         let promiseArray = [];
-        if (entity.isCluster()) {
```
tbouron: Is there really another case, outside a cluster or fabric where we can have a memberspec?

ui-modules/blueprint-composer/app/views/main/yaml/yaml.state.js
ahgittin: It could be useful for things in the YAML editor.  Specifically there is a downstream project which wants this for action enablement.  I could also see it wanted for code completion.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
ahgittin: without this when we `console.log(entity)` we can't identify different entities, and it looks weird, as it just shows `Entity { lastUpdated: xxxx }`.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
ahgittin: we do have the model, maybe not perfect, but we use it already in some places -- eg in composer examine the declared type of a config key to figure out how to render it.  there are some gaps to have this everywhere reliably but feels worth keeping the comment that that would be a better way.

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
ahgittin: yes -- `BrooklynNode` uses it as part of upgrade, test framework uses it i think, and there is a downstream project that wants to use this in a different type of cluster.

tbouron: Ah good point, this should be removed then.

ui-modules/blueprint-composer/app/views/main/yaml/yaml.state.js
ahgittin: have commented out and added comments.  ideally we'd have a cache and then re-enable this but agree with you as it stands it is likely to be too chatty with this enabled.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
ahgittin: okay i'll comment out and we can re-enable if it's wanted for debugging

PR #68, opened 2018-09-11
composer tidies

ahgittin:
* give the right editor for generics in types
* open config section by default
* fix height of dropdown caret in top right


aledsage:
LGTM, merging.

@tbouron is you get a chance to take a look post-merge, that would be appreciated.

PR #69, opened 2018-09-13
Fix a couple issues with composer spec/config editor

ahgittin:
* more on generics
* multiple updates to values / scope variable dereferencing

### DIFF
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
@@ -558,14 +558,6 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             }
             return widgetMetadata.enabled ? "Use standard widget" : "Use custom widget";
         };
-        scope.copyScopeForCustomConfigWidget = (descendantScope) => {
-            descendantScope.toggleCustomConfigWidgetMode = scope.toggleCustomConfigWidgetMode;
-            descendantScope.getCustomConfigWidgetModeTitle = scope.getCustomConfigWidgetModeTitle;
-            descendantScope.defined = scope.defined;
-            descendantScope.config = scope.config;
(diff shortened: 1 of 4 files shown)


aledsage:
FYI without Alex's fix for the custom widget, editing the value behaves really strangely. Below is a description of the behaviour I saw without this fix.

For example, for a dropdown widget, select a value (e.g. “t3.micro”), and then immediately clicking to view the yaml (or click “Deploy"), then the value is not included. However, if I choose the value and then click on another config field, then the value does take effect.

Similarly if I try to type text in there, it goes weird. I had previously chosen “t3.micro” from the dropdown. I then clicked on the text box, deleted the text, and typed “foo”. When I then clicked on another field and then went to the yaml view, the value was “t3.mediu” (when it should have been "foo").

With a new entity, clicking in the box and typing “foo” and then return, the yaml showed just “f”.

Also, the validation error does not indicate the problem (that config is `constraint: [required]`)Instead, the error was only shown at the top-level Configuration.

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -357,6 +357,13 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             return issues.some(issue => issue.level === ISSUE_LEVEL.ERROR) ? 'badge-danger' : 'badge-warning';
         };
 
+        function baseType(s) {
```
aledsage: Do we need to declare `function baseType` here as well, given that we have the import `baseType` from `../util/model/entity.model` as well?

ahgittin:
updated, addressing aled's comment and 1:1 discussion with @tbouron to avoid the use of `$parent` entirely by leaning more on the spec editor's controller, and making a hard requirement on `specEditor`

grkvlt:
Looks good so far. Will test with a custom widget I have been developing to check that it still works...

PR #75, opened 2018-09-26
Composer use controller

ahgittin:
put a few more things on controller for use in custom widgets


ahgittin:
reviewed w @grkvlt ; merging

PR #76, opened 2018-09-28
Composer palette: search entity tags

aledsage:
The palette should also search the entity's tags, when filtering. People can use tags to categorise or provide additional metadata, so this is useful to include.


aledsage:
retest this please

ahgittin:
lgtm

PR #79, opened 2018-10-09
Implement missing endpoint for palette API

tbouron:
The palette API was missing crucial endpoints used to load blueprint from the catalog UI module.

This also fixes a typo in for the `getTypeVersions` method name => `s/getTypeVersion/getTypeVersions`


aledsage:
LGTM; I grep'ed for uses of `getTypeVersion` to check if anything was using the old name - all looks good. Merging.

PR #80, opened 2018-10-09
Add binary requirement to README

tbouron:
This was only notified on the `doc/skinning` docs! I'll make this resurface here. 

We also need this on the Brooklyn docs


tbouron:
Good point @kemitix, I'll add that and merge

PR #82, opened 2018-10-10
Fix image animation for memberspec when there are children to the cluster

tbouron:
Previously, there was an issue on the memberspec node's animation when a cluster had more than one child. This fixes it for both Chrome and Firefox

## Before
screen shot 2018-10-10 at 13 54 16

## After
screen shot 2018-10-10 at 13 53 09


ahgittin:
LGTM

PR #83, opened 2018-10-11
Fix the remove button within overflow menu for adjuncts

tbouron:
Previously, the `Remove` item present within the overflow menu (top right) of the `spec-editor` wasn't removing an adjunct. There was 2 issues causing this behaviour:
1. The `Remove` item wasn't checking the entity family, hence didn't delete any adjunct (as methods are different than removing a child)
2. The `refreshTypeMetadata` was setting the wrong entity family for adjuncts


ahgittin:
LGTM

PR #84, opened 2018-10-11
New layout for the blueprint composer

tbouron:
This allows to:
- add multiple visualisations, if we wish to. Will be added as tabs
- customise palettes. Add/remove sections, they are on the left side
- separate the palette from the configuration screen, less confusing for users
- hide both the palette and spec editor for showing off the blueprint fullscreen

Here is how it looks (Sorry, GitHub does not support video upload in comments)
https://drive.google.com/file/d/1-6b-q6K4mxy3-HSSmRqDLamkVgTliCdZ/view?usp=sharing

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.template.html
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.template.html
@@ -16,4 +16,4 @@
   specific language governing permissions and limitations
   under the License.
 -->
-<button class="btn btn-info" ng-click="activateModal()">{{buttonText}}</button>
+<button class="btn btn-info btn-ouline" ng-click="activateModal()">{{buttonText}}</button>
--- a/ui-modules/blueprint-composer/app/components/designer/designer.less
+++ b/ui-modules/blueprint-composer/app/components/designer/designer.less
@@ -35,7 +35,7 @@ designer {
 
     svg {
         width: 100%;
-        height: calc(~"100vh - 55px");
+        height: 100%;
         cursor: move;
         position: absolute;
         z-index: 1;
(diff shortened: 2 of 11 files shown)


ahgittin:
and it works great.  huge improvement!

one thing, the palette looks odd now not filling the vertical space at left.  can we set it up to do jquery-style math e.g. set the number of items to be `Math.max(16, 4 * ((int)paletteAvailableItems.height / 32))` ?

apart from that, and comments above, good to merge.  (and the palette filling could be pushed to another PR if you're pressed for time.)

tbouron:
@ahgittin Are you talking about the action palette (as in where the items are displayed) or the toolbar on the left to switch between palette sections?

The former actually takes all the vertical space, but as we display only 26 items, it might not look like that. The latter would feel odd IMO if we were to even out the sections, as it has only icons there. Sidebar menus don't tend to do this even though they don't take all vertical space. Do you have any example app that do this?

I'm happy to discuss this point here but you probably right, it can be addressed within another PR.

ahgittin:
the action palette.  it's the limit of 26 (or 24?) that should be replaced by the formula noted above IMO.  i don't know any paginated view that displays fewer items per page than space allows, i think pretty clear that's weird and suboptimal.

tbouron:
@ahgittin This PR does change the behaviour for the `catalog-selector`, it was already doing this before my change. Agree it's not optimal but it get really complicated to check for the actual space available in Javascript, because it depends of the actual screen size, as well as each item size (defined in CSS)

You didn't see this before because the palette was built with multiple accordions, making it bigger than the screen height in almost every occasion

ahgittin:
yes, that's obvious :) -- my point is that this behaviour which was reasonable is now strange and i'd say a defect, and requires some ingenuity to fix.  but can do as another PR.

so two minor changes.  i'll make them and merge.

ahgittin:
scratch the `btn-outlline` suggestion -- it's called `btn-ouline` [sic] everywhere.  so one minor change (required fields).  merging with that.

PR #87, opened 2018-10-22
Allow to resolve different virtual type according to the entity configuration

ludop-usharesoft:
One possible use case of the composer is to have a custom palette that represents several elements ("virutal types") that are based on the same brooklyn type with different brooklyn configuration (referencing different external element like docker image or, in our case, UForge appliance).

In that use case, with the following yaml
```
services:
  - type: com.usharesoft.brooklyn.type.LinuxAppliance
    brooklyn.config:
      applianceUUID: 'AAAA-AAAA-AAAA'
  - type: com.usharesoft.brooklyn.type.LinuxAppliance
    brooklyn.config:
      applianceUUID: 'BBBB-BBBB-BBBB'
  - type: com.usharesoft.brooklyn.type.LinuxAppliance
    brooklyn.config:
      applianceUUID: 'AAAA-AAAA-AAAA'
```
the three services of the same `com.usharesoft.brooklyn.type.LinuxAppliance` brooklyn type should be resolved by two different "virtual" types from the composer point of view, one for the reference `AAAA-AAAA-AAAA` and one for the reference `BBBB-BBBB-BBBB`.

So that graphically we can have when switching from the yaml editor to the graphical editor
image

In order to be able to do that, we need to add an extra third parameters with the entity configuration when the type is resolved by the palette-api provider (when the user switch from the yaml editor to the graphical editor).

In the standard use case of the palette, this extra parameter is ignored and the palette-api simply calls the catalog-api as before.

PS: it's one of my first pull request on the project so don't hesitate to correct me and to tell me the proper way of doing it. Thanks !

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -232,7 +232,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
 
             let promise = entity.miscData.has('bundle')
                 ? paletteApi.getBundleType(entity.miscData.get('bundle').symbolicName, entity.miscData.get('bundle').version, entity.type, entity.version)
-                : paletteApi.getType(entity.type, entity.version);
+                : paletteApi.getType(entity.type, entity.version, entity.config);
 
             promise.then((data)=> {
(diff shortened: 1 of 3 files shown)


ludop-usharesoft:
No build has been triggered on this pull request, probably because I commit it during github issues https://blog.github.com/2018-10-21-october21-incident-report/
I will close this one a create a new one to see if it's change something.
Sorry for the inconvenience.

PR #88, opened 2018-10-22
Allow to resolve different virtual type according to the entity config

ludop-usharesoft:
One possible use case of the composer is to have a custom palette that represents several elements ("virutal types") that are based on the same brooklyn type with different brooklyn configuration (referencing different external element like docker image or, in our case, UForge appliance).

In that use case, with the following yaml
```
services:
  - type: com.usharesoft.brooklyn.type.LinuxAppliance
    brooklyn.config:
      applianceUUID: 'AAAA-AAAA-AAAA'
  - type: com.usharesoft.brooklyn.type.LinuxAppliance
    brooklyn.config:
      applianceUUID: 'BBBB-BBBB-BBBB'
  - type: com.usharesoft.brooklyn.type.LinuxAppliance
    brooklyn.config:
      applianceUUID: 'AAAA-AAAA-AAAA'
```
the three services of the same `com.usharesoft.brooklyn.type.LinuxAppliance` brooklyn type should be resolved by two different "virtual" types from the composer point of view, one for the reference `AAAA-AAAA-AAAA` and one for the reference `BBBB-BBBB-BBBB`.

So that graphically we can have when switching from the yaml editor to the graphical editor
image

In order to be able to do that, we need to add an extra third parameters with the entity configuration when the type is resolved by the palette-api provider (when the user switch from the yaml editor to the graphical editor).

In the standard use case of the palette, this extra parameter is ignored and the palette-api simply calls the catalog-api as before.

PS: it's one of my first pull request on the project so don't hesitate to correct me and to tell me the proper way of doing it. Thanks !

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -232,7 +232,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
 
             let promise = entity.miscData.has('bundle')
                 ? paletteApi.getBundleType(entity.miscData.get('bundle').symbolicName, entity.miscData.get('bundle').version, entity.type, entity.version)
-                : paletteApi.getType(entity.type, entity.version);
+                : paletteApi.getType(entity.type, entity.version, entity.config);
 
             promise.then((data)=> {
                 deferred.resolve(populateEntityFromApiSuccess(entity, data));
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -231,7 +231,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
             entity.family = family;
 
             let promise = entity.miscData.has('bundle')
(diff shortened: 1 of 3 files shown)


ahgittin:
looks good @ludop-usharesoft 

do we have an ICLA on file for you?  good to merge IMO once that's confirmed.

as an aside cc @tbouron i wonder if we should store as UI metadata -- probably in the YAML blueprint -- the virtual type, rather than look it up fresh each time.  not urgent but something to ponder...

ludop-usharesoft:
Hi Alex. 
Yes I have already signed and sent the ICLA to secretary@apache.org (on the 14th of June 2018 with my usharesoft email if you need to search it).
Thanks !

tbouron:
I'm not following your suggestion @ahgittin, could you develop on that please?

ludop-usharesoft:
For the getBundleType method, I have currently use the "bundleVersion" parameter to pass the appliance reference.
So for now it's working without adding the config parameter to it, but yeah it's look like a hack so it's probably not the cleaner way to do that.
However "paletteApi.getBundleType" method is also called from main.controller.js and I don't know exactly in which case.
Maybe it can be a problem not to have the entity.config available there if we modify the getBundleType prototype.

tbouron:
@ludop-usharesoft Just check it it would mean to apply the changes you made but for `getBundleType` on the same files and places. Nothing more would be required.

ludop-usharesoft:
@tbouron ok, here it is.
Thanks !

PR #89, opened 2018-10-24
spec editor hook

frogfather:
Hook to allow downstream to modify the spec editor header


ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -130,7 +130,13 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
                 open: false
             }
         };
-
+        // allow downstream to configure this controller and/or scope
+        (composerOverrides.configureSpecEditor || function() {})(this, $scope, $element);
+        var headerTemplate = angular.element($templateCache.get('spec-editor/header.html'));
```
grkvlt: Can we move this to the callback instead

tbouron:
LGTM, thanks @frogfather

PR #90, opened 2018-10-24
DSL editor: allow ref to entity anywhere

aledsage:
This change means that the DSL editor will let you choose a reference to an entity in any config key (rather than only if the config key is of type `Entity`).

This makes it more consistent with the rest of the DSL editor: e.g. if generating `attributeWhenReady`, it doesn't type-check the attributes' types to filter the view to only those that match the config key's type. This is a good thing, because otherwise we'd have to repeat all the type coercion rules in the UI, for it to know what is legal!

---
I tested this by running `npm run start` in brooklyn-ui/ui-modules/blueprint-composer. I chose a couple of arbitrary entities, and successfully wired up a config key's value to reference the other entity:

dsl-editor-entity-ref

### DIFF
--- a/ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
+++ b/ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
@@ -319,15 +319,13 @@ export function dslEditorDirective($rootScope, $filter, $log, brUtilsGeneral, bl
     function getEntityItems(entity, type) {
         let entities = [];
 
-        if (entity.miscData.get('traits').some(trait => trait.match(type)) || !angular.isDefined(type)) {
-            entities.push({
-                id: entity._id,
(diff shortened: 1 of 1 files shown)


tbouron:
retest this please

(as been aborted because build time > 40min. I tweaked the config to allow for more time)

aledsage:
retest this please

Failed with seemingly unrelated failure in a different module:
```
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.3:npm (npm install) on project brooklyn-ui-location-manager: Failed to run task: 'npm install' failed. java.lang.InterruptedException -> [Help 1]
```

aledsage:
Ah, you beat me to it @tbouron (and I should have refreshed my browser page to see your comment!). The build worked after you requested the retest. My retest was unnecessary.

PR #91, opened 2018-10-24
Add callback for customising catalog save configuration

grkvlt:
Does not include any specific changes for the configuration here, these can be added in future pull requests

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -23,8 +23,11 @@ import template from './catalog-saver.template.html';
 import modalTemplate from './catalog-saver.modal.template.html';
 import jsYaml from 'js-yaml';
 import brUtils from 'brooklyn-ui-utils/utils/general';
+import {yamlState} from "../../views/main/yaml/yaml.state";
+import {graphicalState} from "../../views/main/graphical/graphical.state";
 
 const MODULE_NAME = 'brooklyn.components.catalog-saver';
+
 const REASONS = {
     new: 0,
     deploy: 1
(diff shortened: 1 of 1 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -54,11 +57,17 @@ export function saveToCatalogModalDirective($rootScope, $uibModal) {
         link: link
     };
 
-    function link($scope) {
+    function link($scope, $element) {
         $scope.buttonText = $scope.config.label || ($scope.config.itemType ? `Update ${$scope.config.name || $scope.config.symbolicName}` : 'Add to catalog');
+
+        $injector.get('$templateCache').put('catalog-saver.modal.template.html', modalTemplate);
```
tbouron: You know that you can inject directly `$templateCache` without using the `$injector` service? Might be cleaner as you don't reuse it.

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -54,11 +57,17 @@ export function saveToCatalogModalDirective($rootScope, $uibModal) {
...
+
         $scope.activateModal = () => {
+            // Override callback to update catalog configuration data in other applications
+            $scope.config = (composerOverrides.updateCatalogConfig || (($scope, $element) => $scope.config))($scope, $element);
```
tbouron: I understand why you do this: to augment the scope and/or controller with custom functions/objects.

But I'm wary of auto-calling an anonymous function, especially when the scope is passed around. I don't have a better solution for this however. Any thoughts on the matter @m4rkmckenna ?

PR #92, opened 2018-10-24
Expand relationship support to memberspecs

tbouron:
As the title say, this expand the relationship support to memberspecs, i.e. if a memberspec has a DSL pointing to any other entity, it will now display an arrow to it.

Here is how it looks like:
screen shot 2018-10-24 at 19 52 48

And this done with the classic 3 tier webapp blueprint:
```yaml
name: My Web Cluster
services:
  - type: org.apache.brooklyn.entity.database.mysql.MySqlNode
    id: db
    name: My DB (MySQL)
    brooklyn.config:
      datastore.creation.script.url: >-
        https://raw.githubusercontent.com/apache/brooklyn-library/master/examples/simple-web-cluster/src/main/resources/visitors-creation-script.sql
  - type: org.apache.brooklyn.entity.group.DynamicCluster
    name: My Cluster
    id: cluster
    brooklyn.config:
      initialSize: 2
      dynamiccluster.memberspec:
        '$brooklyn:entitySpec':
          type: org.apache.brooklyn.entity.webapp.tomcat.TomcatServer
          name: Tomcat Server
          brooklyn.config:
            wars.root: >-
              http://search.maven.org/remotecontent?filepath=org/apache/brooklyn/example/brooklyn-example-hello-world-sql-webapp/0.9.0/brooklyn-example-hello-world-sql-webapp-0.9.0.war
            java.sysprops:
              brooklyn.example.db.url: >-
                $brooklyn:formatString("jdbc:%s%s?user=%s&password=%s",
                $brooklyn:component("db").attributeWhenReady("datastore.url"), "visitors", "brooklyn", "br00k11n")
    brooklyn.enrichers:
      - type: org.apache.brooklyn.enricher.stock.Aggregator
        brooklyn.config:
          enricher.sourceSensor: '$brooklyn:sensor("webapp.reqs.perSec.windowed")'
          enricher.targetSensor: '$brooklyn:sensor("webapp.reqs.perSec.perNode")'
          enricher.aggregating.fromMembers: true
          transformation: average
  - type: org.apache.brooklyn.entity.proxy.nginx.NginxController
    id: nginx
    name: My Load Balancer (nginx)
    brooklyn.config:
      loadbalancer.serverpool: '$brooklyn:entity("cluster")'
      nginx.sticky: false
brooklyn.enrichers:
  - type: org.apache.brooklyn.enricher.stock.Propagator
    brooklyn.config:
      producer: '$brooklyn:entity("nginx")'
      propagating:
        - main.uri
        - main.uri.mapped.subnet
        - main.uri.mapped.public

```

Poke @ahgittin

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -625,6 +625,13 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
             };
         });
 
+        relationships = Array.from(entity.config.values())
+            .filter(config => config[DSL_ENTITY_SPEC] && config[DSL_ENTITY_SPEC] instanceof Entity)
+            .map(config => config[DSL_ENTITY_SPEC])
+            .reduce((relationships, spec) => {
+            return relationships.concat(getRelationships(spec));
+        }, relationships);
+
         return entity.children.reduce((relationships, child) => {
             return relationships.concat(getRelationships(child))
         }, relationships);
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -677,7 +677,13 @@ export function D3Blueprint(container) {
      * @return {*} a D3 tree node
      */
     function nodeForEntity(entity) {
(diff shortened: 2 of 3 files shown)


ahgittin:
LGTM.  feels like we should support multiple spec config keys in which case the single field `memberspec` wouldn't be quite right and we'd need to compute a different `x`  as we do for `y` -- but not high priority, not needed here

PR #93, opened 2018-10-25
new layout (separable palette) tidies

ahgittin:
#84 introduces the separable palette/config panel

this enhances that so that the palette gets the right size, and also fixes extension points and makes it so the "add (member spec)" usage of the panel respects those extension points

minor tidies to dropdowns and other style points also

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
@@ -39,6 +39,7 @@ export function catalogSelectorDirective() {
             onSelect: '&',
             rowsPerPage: '<',  // if unset then fill
             reservedKeys: '<?',
+            state: '<?',
             mode: '@?',  // for use by downstream projects to pass in special modes
         },
         template: template,
@@ -147,15 +148,15 @@ function controller($scope, $element, $timeout, $q, $uibModal, $log, $templateCa
     this.$timeout = $timeout;
 
     $scope.viewModes = PALETTE_VIEW_MODES;
-    $scope.state = {
-        orders: ['name', 'type', 'id'],
-        currentOrder: 'name',
-        viewMode: PALETTE_VIEW_MODES.normal,
-    };
+    $scope.viewOrders = ['name', 'type', 'id'];
+    
+    if (!$scope.state) $scope.state = {};
+    if (!$scope.state.viewMode) $scope.state.viewMode = PALETTE_VIEW_MODES.normal;
+    if (!$scope.state.currentOrder) $scope.state.currentOrder = 'name';
     
     $scope.pagination = {
         page: 1,
-        itemsPerPage: $scope.state.viewMode.itemsPerRow * ($scope.rowsPerPage || MIN_ROWS_PER_PAGE)
+        itemsPerPage: $scope.state.viewMode.itemsPerRow * ($scope.rowsPerPage || 1)  // will fill out after load
     };
     
     $scope.getEntityNameForPalette = function(item, entityName) {
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
@@ -20,7 +20,14 @@ import angular from 'angular';
 import {EntityFamily} from '../util/model/entity.model';
 import template from './catalog-selector.template.html';
 
-const MIN_ITEMS_PER_PAGE = 16;
+const MIN_ROWS_PER_PAGE = 4;
+
+const PALETTE_VIEW_MODES = {
+  compact: { name: "Compact", classes: "col-xs-2 item-compact", itemsPerRow: 6, rowHeightPx: 75, hideName: true },
+  normal: { name: "Normal", classes: "col-xs-3", itemsPerRow: 4 },
+  large: { name: "Large", classes: "col-xs-4", itemsPerRow: 3 },
+  list: { name: "List", classes: "col-xs-12 item-full-width", itemsPerRow: 1 } };
+
 // fields in either bundle or type record:
 const FIELDS_TO_SEARCH = ['name', 'displayName', 'symbolicName', 'version', 'type', 'supertypes', 'containingBundle', 'description', 'displayTags', 'tags'];
 
@@ -30,7 +37,7 @@ export function catalogSelectorDirective() {
         scope: {
             family: '<',
             onSelect: '&',
-            itemsPerPage: '<',
+            rowsPerPage: '<',  // if unset then fill
             reservedKeys: '<?',
             mode: '@?',  // for use by downstream projects to pass in special modes
         },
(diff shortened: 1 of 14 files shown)


ahgittin:
retest this please

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -108,15 +144,21 @@ export function catalogSelectorSortFilter($filter) {
...
-function controller($scope, $element, $q, $uibModal, $log, $templateCache, paletteApi, paletteDragAndDropService, iconGenerator, composerOverrides) {
+function controller($scope, $element, $timeout, $q, $uibModal, $log, $templateCache, paletteApi, paletteDragAndDropService, iconGenerator, composerOverrides) {
+    this.$timeout = $timeout;
+
+    $scope.viewModes = PALETTE_VIEW_MODES;
+    $scope.viewOrders = ['name', 'type', 'id'];
+    
+    if (!$scope.state) $scope.state = {};
+    if (!$scope.state.viewMode) $scope.state.viewMode = PALETTE_VIEW_MODES.normal;
+    if (!$scope.state.currentOrder) $scope.state.currentOrder = 'name';
```
tbouron: Could you use curly brackets for the `if` statement please? That's to be consistent with the rest of the codebase and also make it easier to read

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -127,15 +169,15 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
...
+    
     $scope.isLoading = true;
 
     $scope.$watch('search', () => {
         $scope.freeFormTile = {
             symbolicName: $scope.search,
             name: $scope.search,
             displayName: $scope.search,
-            supertypes: [$scope.family.superType]
+            supertypes: ($scope.family ? [ $scope.family.superType ] : []),
```
tbouron: `$scope.family` is marked as required so this should not be needed. Callers of the `catalog-selector` directive should always pass it

ui-modules/blueprint-composer/app/components/providers/palette-service.provider.js
```diff
@@ -37,7 +37,9 @@ export function paletteServiceProvider() {
             sections[id] = section;
         },
         deleteSection(id) {
+            let old = sections[id];
             delete sections[id];
+            return old;
```
tbouron: I'm curious, why returning `old`?

ui-modules/blueprint-composer/app/views/main/graphical/edit/add/add.js
```diff
@@ -51,8 +52,11 @@ export function GraphicalEditAddController($scope, $filter, $state, $stateParams
             break;
     }
 
-    $scope.catalogItemsPerPage = 24;
-
+    if (!$scope.familiesToShow) $scope.familiesToShow = [ $scope.family ];
```
tbouron: Could you use curly brackets for the `if` statement please? That's to be consistent with the rest of the codebase and also make it easier to read

ui-modules/blueprint-composer/app/views/main/graphical/edit/add/add.js
```diff
@@ -27,17 +27,18 @@ export const graphicalEditAddState = {
...
 
-export function GraphicalEditAddController($scope, $filter, $state, $stateParams, blueprintService) {
+export function GraphicalEditAddController($scope, $filter, $state, $stateParams, blueprintService, paletteService) {
     switch ($stateParams.family) {
         case EntityFamily.ENTITY.id.toLowerCase():
             $scope.family = EntityFamily.ENTITY;
             break;
         case EntityFamily.SPEC.id.toLowerCase():
             $scope.family = EntityFamily.SPEC;
+            $scope.familiesToShow = [ EntityFamily.ENTITY, EntityFamily.SPEC ];
```
tbouron: Seems to be used only in the JS code so would be better to be a local var (i.e. avoiding adding stuff to the `$scope` as a general rule of thumb)

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -243,4 +285,6 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
 
     // allow downstream to configure this controller and/or scope
     (composerOverrides.configurePaletteController || function() {})(this, $scope, $element);
+    
+    $scope.paletteItemClasses = "col-xs-3";
```
tbouron: This doesn't seems to be used so can be removed

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -20,7 +20,14 @@ import angular from 'angular';
 import {EntityFamily} from '../util/model/entity.model';
 import template from './catalog-selector.template.html';
 
-const ITEMS_PER_PAGE = 30;
+const MIN_ROWS_PER_PAGE = 4;
+
+const PALETTE_VIEW_MODES = {
```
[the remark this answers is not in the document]
ahgittin: not convinced.  could be changed when useful.

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
ahgittin: can be passed in as null however

ui-modules/blueprint-composer/app/components/providers/palette-service.provider.js
ahgittin: if you want to change order a common pattern would be to delete it but put it to a local var so you could then add it in the right place.  same idea as `Map.remove`

ui-modules/blueprint-composer/app/views/main/graphical/edit/add/add.js
ahgittin: no, it's used in the html

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
ahgittin: yes, removed in #94

ahgittin:
all addressed as part of #93

ui-modules/blueprint-composer/app/views/main/graphical/edit/add/add.js
tbouron: I cannot find any reference if this on templates

ui-modules/blueprint-composer/app/views/main/graphical/edit/add/add.js
tbouron: I cannot find any reference if this on templates

ahgittin: https://github.com/apache/brooklyn-ui/pull/93/files#diff-ed969f9a529823af9b9282a89cc5ebc1R31

tbouron:
Thanks @ahgittin, merging

PR #94, opened 2018-10-26
Composer recent and filters

ahgittin:
store how recently an entity or palette icon was used in the composer, and introduce a filter and sort to use this info.  also move sorts and dropdown to new places as shown below.  fix sort (was not working), allow multi-sort (just remembers the last things you sorted from, does not show them), and add a footer to describe the results of sorts and filters

builds on #93 so review that first (or review the two togethers

screen shot 2018-10-26 at 12 15 19

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
@@ -1,3 +1,21 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
 <div class="paletteItemsFooter">
     <div ng-if="searchedItems.length === 0" class="no-match">
         <i class="fa fa-lemon-o"></i>
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
@@ -1,0 +1,42 @@
+<div class="paletteItemsFooter">
+    <div ng-if="searchedItems.length === 0" class="no-match">
+        <i class="fa fa-lemon-o"></i>
+    </div>
+    <small class="help-block text-sm palette-footer-message">
+        <span ng-if="search">
+          <span ng-if="searchedItems.length === 0">
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    matching search but excluded by filters. 
+                    <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
+            </span>
+            <span ng-if="skippingFilters || itemsBeforeActiveFilters.length === 0">
+                    No matches for <code>{{ search }}</code>.
+            </span>
+          </span>
+          <span ng-if="searchedItems.length > 0">
+            <span ng-if="skippingFilters">
+                    <b>No matches with selected filters.</b><br/>
+                    Showing matches ignoring filters.
+            </span>
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    matching search but excluded by filters.
+                    <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
+            </span>
+          </span>
+        </span>
+        <span ng-if="!search">
+            <span ng-if="skippingFilters && searchedItems.length > 0">
+                    <b>Nothing available in selected filters.</b><br/>
+                    Ignoring filters.
+            </span>
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    excluded by filters. 
+                    <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
+            </span>
+        </span>
+        
+    </small>
+</div>
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
@@ -17,14 +17,15 @@
   under the License.
 -->
 <div class="paletteItemsFooter">
-    <div ng-if="searchedItems.length === 0" class="no-match">
-        <i class="fa fa-lemon-o"></i>
+    <div ng-if="searchedItems.length === 0" class="icon no-match">
+        <span>&oslash;</span>
     </div>
-    <small class="help-block text-sm palette-footer-message">
+    <small class="help-block text-sm palette-footer-message" ng-class="{ 'no-match': searchedItems.length === 0 }">
         <span ng-if="search">
           <span ng-if="searchedItems.length === 0">
             <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
-                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    <strong>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} 
+                        <ng-pluralize count="itemsBeforeActiveFilters.length" when="{ '1': 'item', 'other': 'items' }"></<ng-pluralize></strong> 
                     matching search but excluded by filters. 
                     <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
             </span>
@@ -34,23 +35,28 @@
           </span>
           <span ng-if="searchedItems.length > 0">
             <span ng-if="skippingFilters">
-                    <b>No matches with selected filters.</b><br/>
+                    <strong>No matches with selected filters.</strong><br/>
                     Showing matches ignoring filters.
             </span>
             <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
-                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    <strong>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more 
+                        <ng-pluralize count="itemsBeforeActiveFilters.length" when="{ '1': 'item', 'other': 'items' }"></<ng-pluralize></strong> 
                     matching search but excluded by filters.
                     <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
             </span>
           </span>
         </span>
         <span ng-if="!search">
+            <span ng-if="itemsBeforeActiveFilters.length == 0">
+                <strong>Nothing available.</strong>
+            </span>
             <span ng-if="skippingFilters && searchedItems.length > 0">
-                    <b>Nothing available in selected filters.</b><br/>
+                    <strong>Nothing available in selected filters.</strong><br/>
                     Ignoring filters.
             </span>
             <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
-                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    <strong>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more 
+                        <ng-pluralize count="itemsBeforeActiveFilters.length" when="{ '1': 'item', 'other': 'items' }"></<ng-pluralize></strong> 
                     excluded by filters. 
                     <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
             </span>
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
@@ -39,6 +39,7 @@ export function catalogSelectorDirective() {
             onSelect: '&',
             rowsPerPage: '<',  // if unset then fill
             reservedKeys: '<?',
+            state: '<?',
             mode: '@?',  // for use by downstream projects to pass in special modes
         },
         template: template,
@@ -147,15 +148,15 @@ function controller($scope, $element, $timeout, $q, $uibModal, $log, $templateCa
     this.$timeout = $timeout;
 
     $scope.viewModes = PALETTE_VIEW_MODES;
-    $scope.state = {
-        orders: ['name', 'type', 'id'],
-        currentOrder: 'name',
-        viewMode: PALETTE_VIEW_MODES.normal,
-    };
+    $scope.viewOrders = ['name', 'type', 'id'];
+    
+    if (!$scope.state) $scope.state = {};
+    if (!$scope.state.viewMode) $scope.state.viewMode = PALETTE_VIEW_MODES.normal;
+    if (!$scope.state.currentOrder) $scope.state.currentOrder = 'name';
     
     $scope.pagination = {
         page: 1,
-        itemsPerPage: $scope.state.viewMode.itemsPerRow * ($scope.rowsPerPage || MIN_ROWS_PER_PAGE)
+        itemsPerPage: $scope.state.viewMode.itemsPerRow * ($scope.rowsPerPage || 1)  // will fill out after load
     };
     
     $scope.getEntityNameForPalette = function(item, entityName) {
--- a/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
@@ -19,6 +19,7 @@
 import angular from 'angular';
 import {EntityFamily} from '../util/model/entity.model';
 import template from './catalog-selector.template.html';
+import footerTemplate from './catalog-selector-palette-footer.html';
 import moment from "moment";
 
 const MIN_ROWS_PER_PAGE = 4;
@@ -281,7 +282,10 @@ function controller($scope, $element, $timeout, $q, $uibModal, $log, $templateCa
                 return $scope.recentItems; 
             }, enabled: false },
     ];
-    $scope.disableFilters = () => $scope.filters.forEach( f => f.enabled = false );
+    $scope.disableFilters = (showFilters) => {
+        $scope.filters.forEach( f => f.enabled = false );
+        if (showFilters !== false) $scope.showPaletteControls = true;
+    }
     
     // this can be overridden for palette sections/modes which show a subset of the types returned by the server;
     // this is applied when the data is received from the server.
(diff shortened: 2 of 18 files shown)


ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
```diff
@@ -0,0 +1,60 @@
...
+-->
+<div class="paletteItemsFooter">
+    <div ng-if="searchedItems.length === 0" class="no-match">
+        <i class="fa fa-lemon-o"></i>
+    </div>
+    <small class="help-block text-sm palette-footer-message">
+        <span ng-if="search">
+          <span ng-if="searchedItems.length === 0">
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b>
```
tbouron: Should use `<strong>` instead of `<b>` tag. Also, should use `ng-pluralize` (https://docs.angularjs.org/api/ng/directive/ngPluralize) to handle plurals

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
```diff
@@ -0,0 +1,60 @@
...
+                    matching search but excluded by filters. 
+                    <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
+            </span>
+            <span ng-if="skippingFilters || itemsBeforeActiveFilters.length === 0">
+                    No matches for <code>{{ search }}</code>.
+            </span>
+          </span>
+          <span ng-if="searchedItems.length > 0">
+            <span ng-if="skippingFilters">
+                    <b>No matches with selected filters.</b><br/>
```
tbouron: Should use `<strong>` instead of `<b>` tag.
```suggestion
                    <strong>No matches with selected filters.</strong><br/>
```

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
```diff
@@ -0,0 +1,60 @@
...
+                    Showing matches ignoring filters.
+            </span>
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b>
```
tbouron: Same as 2 comments above

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
```diff
@@ -0,0 +1,60 @@
...
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b> 
+                    matching search but excluded by filters.
+                    <button class="btn btn-outline btn-info" ng-click="disableFilters()">Clear filters</button>
+            </span>
+          </span>
+        </span>
+        <span ng-if="!search">
+            <span ng-if="skippingFilters && searchedItems.length > 0">
+                    <b>Nothing available in selected filters.</b><br/>
```
tbouron: Should use `<strong>` instead of `<b>` tag.
```suggestion
                    <strong>Nothing available in selected filters.</strong><br/>
```

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
```diff
@@ -0,0 +1,60 @@
...
+                    Ignoring filters.
+            </span>
+            <span ng-if="!skippingFilters && itemsBeforeActiveFilters.length > itemsAfterActiveFilters.length">
+                    <b>{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length }} more item{{ itemsBeforeActiveFilters.length - itemsAfterActiveFilters.length > 1 ? 's' : ''}}</b>
```
tbouron: Same as 2 comments above

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -30,15 +47,43 @@ export function catalogSelectorDirective() {
         scope: {
             family: '<',
             onSelect: '&',
-            itemsPerPage: '<',
+            rowsPerPage: '<',  // if unset then fill
```
tbouron: If this parameter is option, it should have a `?` at the end, like the ones below

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -30,15 +47,43 @@ export function catalogSelectorDirective() {
...
     };
 }
 
+function link($scope, $element, attrs, controller) {
+    let main = angular.element($element[0].querySelector(".catalog-palette-main"));
+
+    // repaginate when load completes (and items are shown), or it is resized
+    $scope.$watchGroup(
+        [ () => $scope.isLoading, () => main[0].offsetHeight, () => $scope.state.viewMode.itemsPerRow ],
+        (values) => controller.$timeout( () => repaginate($scope, $element) ) );
```
tbouron: Instead of using `$timeout`, you should use `$scope.$applyAsync`

```suggestion
        (values) => $scope.$applyAsync( () => repaginate($scope, $element) ) );
```

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -30,15 +47,43 @@ export function catalogSelectorDirective() {
...
+    // also repaginate on window resize    
+    angular.element(window).bind('resize', () => repaginate($scope, $element));
+}
+
+function repaginate($scope, $element) {
+    let rowsPerPage = $scope.rowsPerPage;
+    if (!rowsPerPage) {
+        let main = angular.element($element[0].querySelector(".catalog-palette-main"));
+        if (!main || main[0].offsetHeight==0) {
+            // console.log("no main or hidden or items per page fixed");
```
tbouron: Either removing this or use the `$log` service

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -65,58 +110,21 @@ export function catalogSelectorSearchFilter() {
...
+function controller($scope, $element, $timeout, $q, $uibModal, $log, $templateCache, paletteApi, paletteDragAndDropService, iconGenerator, composerOverrides, recentlyUsedService) {
+    this.$timeout = $timeout;
 
-function controller($scope, $element, $q, $uibModal, $log, $templateCache, paletteApi, paletteDragAndDropService, iconGenerator, composerOverrides) {
+    $scope.viewModes = PALETTE_VIEW_MODES;
+    $scope.viewOrders = PALETTE_VIEW_ORDERS;
+    
+    if (!$scope.state) $scope.state = {};
+    if (!$scope.state.viewMode) $scope.state.viewMode = PALETTE_VIEW_MODES.normal;
+    if (!$scope.state.currentOrder) $scope.state.currentOrder = [ PALETTE_VIEW_ORDERS.name.field, '-version' ];
```
tbouron: Could you use curly brackets for the `if` statements in order to be consistent with the codebase please? It also makes it easier to read

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -127,15 +135,15 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
...
+    
     $scope.isLoading = true;
 
     $scope.$watch('search', () => {
         $scope.freeFormTile = {
             symbolicName: $scope.search,
             name: $scope.search,
             displayName: $scope.search,
-            supertypes: [$scope.family.superType]
+            supertypes: ($scope.family ? [ $scope.family.superType ] : []),
```
tbouron: `scope.family` is marked as required in the directive specification, you shouldn't need that

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -230,16 +248,85 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
         });
         $scope.items = items;
     });
-    // this can be overridden for third-party filters.
-    // it receives result of filtering based on search so filters can adjust based on number of search resullts
-    $scope.filterPaletteItems = (items) => items;
+    $scope.lastUsedText = (item) => {
+        let l = (Number)(item.lastUsed);
+        if (!l || isNaN(l) || l<=0) return "";
+        if (l < 100000) return 'Preselected for inclusion in "Recent" filter.';
```
tbouron: Could you use curly brackets for the `if` statements in order to be consistent with the codebase please? It also makes it easier to read

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -230,16 +248,85 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
...
+        return 'Last used: ' + moment(l).fromNow();
+    }; 
+    $scope.showPaletteControls = false;
+    $scope.onFiltersShown = () => {
+      $timeout( () => {
```
tbouron: You can use `$scope.$applyAsync` instead, which is cleaner than a timeout
```suggestion
      $scope.$applyAsync( () => {
```

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -230,16 +248,85 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
...
+        // check do we need to show the multiline
+        let filters = angular.element($element[0].querySelector(".filters"));
+        $scope.$apply( () => $scope.filterSettings.filtersMultilineAvailable = filters[0].scrollHeight > filters[0].offsetHeight );
+        
+        repaginate($scope, $element);
+      } );
+    };
+    $scope.togglePaletteControls = () => {
+        $scope.showPaletteControls = !$scope.showPaletteControls;
+        $timeout( () => repaginate($scope, $element) );
```
tbouron: Same as above

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -230,16 +248,85 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
...
+    }
+    $scope.toggleShowAllFilters = () => {
+        $scope.filterSettings.showAllFilters = !$scope.filterSettings.showAllFilters;
+        $timeout( () => repaginate($scope, $element) );
```
tbouron: Same as above

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -230,16 +248,85 @@ function controller($scope, $element, $q, $uibModal, $log, $templateCache, palet
...
+        { label: 'Recent', icon: 'clock-o', title: "Recently used and standard favorites", limitToOnePage: true,
+            filterInit: items => {
+                $scope.recentItems = items.filter( i => i.lastUsed && i.lastUsed > 0 );
+                $scope.recentItems.sort( (a,b) => b.lastUsed - a.lastUsed );
+                return $scope.recentItems; 
+            }, enabled: false },
+    ];
+    $scope.disableFilters = (showFilters) => {
+        $scope.filters.forEach( f => f.enabled = false );
+        if (showFilters !== false) $scope.showPaletteControls = true;
```
tbouron: Could you use curly brackets for the `if` statements in order to be consistent with the codebase please? It also makes it easier to read

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.template.html
```diff
@@ -30,44 +30,83 @@
...
+      </div>
 
-        <div class="row grid catalog-palette">
+        <div class="row grid">
             <!-- here and below, col-xs-3 or -4 or -2 all work giving different densities;
                  this could be configurable ("compressed"=xs-2 w no labels, "normal"=xs-3, "big"=xs-4) -->
-            <div class="col-xs-3 catalog-palette-item"
-                    ng-repeat="item in searchedItems = (filterPaletteItems(items | catalogSelectorSearch:search) | catalogSelectorSort:family) | orderBy:state.currentOrder | limitTo:pagination.itemsPerPage:(pagination.page-1)*pagination.itemsPerPage track by (item.containingBundle + ':' + item.symbolicName + ':' + item.version)"
+            <div class="catalog-palette-item" ng-class="state.viewMode.classes"
+                    ng-repeat="item in searchedItems = filterPaletteItemsWithActiveFilters(items | catalogSelectorSearch:search) | orderBy:state.currentOrder | limitTo:pagination.itemsPerPage:(pagination.page-1)*pagination.itemsPerPage track by (item.containingBundle + ':' + item.symbolicName + ':' + item.version)"
```
tbouron: It's a very bad angular practice to call for a function that returns a new array on every call: the track cannot run properly as it will be new items every time. Considering the number of items in the list, this should use an angular filter instead.

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
```diff
@@ -0,0 +1,62 @@
...
+    
+    service.markUsed = (item, when) => {
+        let id = service.getId(item);
+        if (when) {
+            let old = service.getLastUsed(id);
+            if (old > when) return;
+        } else {
+            when = Date.now();
+        }
+        sessionStorage.setItem(PREFIX+id, when);
```
tbouron: Storing each item in the `sessionStorage` can lead to crippled instances as there is a limited number of items you can store them. Instead, a good solution is to store the entire `Map` of items as a `JSON.stringify()` into one `sessionStorage.setItem()`

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
```diff
@@ -0,0 +1,62 @@
...
+        let id = service.getId(item); 
+        let s = sessionStorage.getItem(PREFIX+id);
+        if (s) return ((Number)(s));
+        let tag = item.tags && item.tags.find(t => t['ui-composer-recent-preselect']);
+        if (tag) return tag['ui-composer-recent-preselect']; 
+        return -1; 
+    };
+    
+    service.embellish = (item) => {
+        item.lastUsed = service.getLastUsed(item);
```
tbouron: This should be added into the modal `miscData`
```suggestion
        item.miscData.set('lastUsed, service.getLastUsed(item));
```

ui-modules/blueprint-composer/package.json
```diff
@@ -60,7 +60,8 @@
     "lodash": "^4.16.4",
     "ngclipboard": "^1.1.1",
     "normalizr": "^2.2.1",
-    "underscore": "^1.8.3"
+    "underscore": "^1.8.3",
+    "moment": "^2.15.1"
```
tbouron: This will increase the bundle size by about 350k. There are smaller alternatives available if you just want to display the time from now like https://date-fns.org/

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.template.html
ahgittin: noted - will work on this

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
ahgittin: hmm, i thought stringifying a potentially very large map on every change would be less performant than using session storage.  do you have more info on the limits?  (the limits i've seen are on _size_ not number of entries.)

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
ahgittin: will do, though curious, why?

ui-modules/blueprint-composer/package.json
ahgittin: nice, will try this

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
tbouron: We did hit the maximum size on the earlier day of the composer while saving the generated icons. Now, I think it's the actual size of the `localStorage` (5MB if I recall correctly) but don't remember if there is a limit per item or in total.

Nonetheless, we need to handle this because `sessionStorage.setItem()` will throw if there is no more space left.

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
tbouron: Mainly for consistency and clear separation of concerns: top fields are the essential information of the model. Secondary one are all put into the `miscData` map

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector-palette-footer.html
ahgittin: ack

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
ahgittin: seems like the issue is that it can through; will handle that with `try ... catch`, good point.  won't convert to stringified map unless you know a good reason to do that.

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
ahgittin: `miscData` is not available here.  i think it's used on things in the blueprint but not things in the catalog?

ui-modules/blueprint-composer/package.json
ahgittin: works a treat

ui-modules/blueprint-composer/app/components/providers/recently-used-service.provider.js
tbouron: Aaah `item` refers here as items from the API, not the model. 
Right so it makes sense then.

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
```diff
@@ -110,6 +110,38 @@ export function catalogSelectorSearchFilter() {
     }
 }
 
+export function catalogSelectorFiltersFilter() {
+    // compute counts and apply active filters;     
+    // this is called by the view after filtering based on search,
+    // so filters can adjust based on number of search results
+    return function (items, $scope) {
```
tbouron: Passing scope around is a bad idea for performances, you should pass each individual parameter, or an object containing all them instead.

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
tbouron: Should use `$scope.$applyAsync` instead of `$timeout`

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
tbouron: Should use `$scope.$applyAsync` instead of `$timeout`

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
tbouron: This parameter is optional, it should have a `?` at the end, like the ones below.

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
tbouron: This directive wasn't designed to support `null` family. Something is wrong is that's the case, I don't see any reason why we would use it without a family

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
tbouron: Should use `$scope.$applyAsync` instead of `$timeout`

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
tbouron: Should use `$scope.$applyAsync` instead of `$timeout`

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
ahgittin: we set variables on the scope when the filter runs.  i take your point a nested object would be cleaner but let's do performance analysis before optimizing.  this doesn't seem to be that inefficient.

ahgittin:
PS i tried with `$scope.$applyAsync` but it ran before the html was reprocessed -- it had `offsetHeight==0` whereas with `$timeout` it is correctly set as `24`.

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
ahgittin: https://stackoverflow.com/questions/17596246/access-scope-variables-from-a-filter-in-angularjs gives good caveats on this

ui-modules/blueprint-composer/app/components/catalog-selector/catalog-selector.directive.js
ahgittin: pretty sure i added it because of an NPE here.  not sure how/why but this check won't hurt.

tbouron:
Ah makes sense, you want the height therefore the template must be compiled. `$scope.$applyAsync` is executing at the beginning of a digest cycle therefore this info is not available.

LGTM then, merging this. Thanks @ahgittin

PR #95, opened 2018-10-31
Add callback for customising yaml loading

sferot:


ahgittin:
good refactor @sferot .  i note you have an ICLA so this is good to go.

heads up, @tbouron has been spiking angularjs decorators which should provide an easier pattern to override/customize.  if you are wanting to override this you could try that and let us know how it goes!

merging

ahgittin:
actually this is giving errors when i run it:

```
TypeError: this.$q.defer is not a function
    at BlueprintLoaderApiProvider.loadBlueprint (blueprint-loader-api.provider.js:51)
```

will try to fix but if i can't i'll have to revert

ahgittin:
@sferot fix pushed, but pls check it's what you intended

https://github.com/apache/brooklyn-ui/commit/65d37408778365d0d987a46b119deea12f0c70d7

ludop-usharesoft:
Thanks @ahgittin for the merge and the fix
@sferot is in holidays since last wednesday, he will come back on the 12th.
If I have the time week I may try to look at it (but not sure).

PR #97, opened 2018-11-07
Generalise usage of $templateCache for blueprint composer

tbouron:
This updated all directives within the blueprint composer to use `$templateCache` service for their templates. It allows downstream projects to update templates really easily with an angular `run` block.

Fo this to be consistent with the rest of the codebase, I also transform each directive to be an angular module: it keeps everything neatly separated and self contained.


aledsage:
Description sounds sensible to me, and scanning through the code I didn't see anything obviously wrong!

I've tested it in `npm run start` mode, and all the relevant blueprint-composer features seem to still work, so I think the restructuring/wiring has been done correctly.

Happy for this to be merged, but if anyone else is reviewing the code more thoroughly then even happier to wait for them!

PR #98, opened 2018-11-07
Add compact list view mode

jcabrerizo:
Added new compact list view mode to the palette with small icons and the full description
image


jcabrerizo:
Yes, I did it a couple of weeks ago

ahgittin:
cool, just that fix then.  and, i've just noticed, this has conflicts with #97  which i've merged.  should be easy to resolve, i think it's the fix to the watch group you added that needs to be reapplied.

jcabrerizo:
Conflicts resolved

PR #99, opened 2018-11-08
make palette icons have a light border not a shadow

ahgittin:
gives it a nicer flat look


tbouron:
The box shadow is there to indicate a notion of hierarchy. The side panels come on top of the canvas, not beside it, i.e. they don't push the canvas to left or right.
Material design says that in this particular case, the panel should be "elevated", same as a piece of paper on top of another one. This creates a shadow, reflected by the box shadow there currently is.

Brooklyn UI was created with material design in mind. It is poorly executed in a lot of places but still. I think it make sense to keep this notion of hierarchy as it fit with what it's doing

tbouron:
@ahgittin Ah I didn't realise you were talking about the rectangles!

I get your point and it is valid. Or rather it was valid until https://github.com/apache/brooklyn-ui/pull/96. Items are now added to the canvas only via drag & drop. This is the primary action, displaying the popover on click is a secondary action for convenience.

As such, we should keep the rectangles elevated to convey that they are a draggable items

ahgittin:
good point although #96 also allows clicking then you get an "add" button.  also there are things in MD which are draggable but not elevated; elevation implies more than draggabililty but some importance to the thing being elevated which isn't wanted here.  so the elevation makes it busier which negates the cognitive hint that it is something you should drag, the thin line is easier on the eye and easier to work with.  but it is a subjective aesthetic decision -- which do you prefer (1) visually, and (2) according to the dogma of MD?

for me, (1) is strong pref for thin line; (2) is on the fence, so thin line wins.

ahgittin:
I get that Thomas, but the squares of paper when resting on the table don't cast a shadow, at least not until you pick it up.  This is how the new Google Mail does it with the list of emails -- they get the shadow when you hover as though that's when you are dragging it.

How about a compromise with a subtle shadow when resting (PR updated):

screenshot 2018-11-21 at 10 28 11

tbouron:
Sounds good. Thank you @ahgittin, I'm happy with that

Merging now

PR #100, opened 2018-11-08
Landscape view - initial refactor

ahgittin:
This refactors the blueprint code to allow dev work on a new "landscape" graphical mode.

It is disabled in the last commit so should be no change to UX with this,
but the other changes should make it easier to work on a landscape view by playing with the sizes of nodes.  They refactor much more of the layout-specific logic into subclasses, make the elements being rendered more flexible, and make more of the computations derived from known values.

There is more work to make things like ghost nodes fully derived from the nodes they ghost, which we will need for the landscape view to be able to fill the space but this gets us a good bit of the way there.

If landscape is enabled, it doesn't do what we want, but it has slightly different boxier behaviour to the normal mode to indicate the direction of travel.

visual1


nakomis:
retest this please

nakomis:
@ahgittin I've tried to build this locally, but I'm getting build errors, and the build is also failing on Jenkins. Can you rebase against master please

PR #106, opened 2018-11-09
Update save to catalog action

grkvlt:
Sets some default values from the blueprint and moves some items to an advanced section

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -42,20 +42,20 @@ const TYPES = [
 ];
 
 angular.module(MODULE_NAME, [angularAnimate, uibModal, brUtils])
-    .directive('catalogSaver', ['$rootScope', '$uibModal', '$injector', 'composerOverrides', saveToCatalogModalDirective])
+    .directive('catalogSaver', ['$rootScope', '$uibModal', '$injector', 'composerOverrides', 'blueprintService', saveToCatalogModalDirective])
     .directive('catalogVersion', ['$parse', catalogVersionDirective])
     .run(['$templateCache', templateCache]);
 
 export default MODULE_NAME;
 
-export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, composerOverrides) {
+export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, composerOverrides, blueprintService) {
     return {
         restrict: 'E',
         templateUrl: function (tElement, tAttrs) {
             return tAttrs.templateUrl || TEMPLATE_URL;
         },
         scope: {
-            config: '='
+            config: '=',
         },
         link: link
     };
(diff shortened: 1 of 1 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -60,12 +60,47 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, co
...
 
-    function link($scope, $element) {
+    function link($scope, $element, $compile, controller) {
         $scope.buttonText = $scope.config.label || ($scope.config.itemType ? `Update ${$scope.config.name || $scope.config.symbolicName}` : 'Add to catalog');
 
         $scope.activateModal = () => {
-            // Override callback to update catalog configuration data in other applications
-            $scope.config = (composerOverrides.updateCatalogConfig || (($scope, $element) => $scope.config))($scope, $element);
+            function injectorGet(reference) { return $element.injector().get(reference); }
+            function blueprintService() { return injectorGet('blueprintService'); }
```
tbouron: Instead of using the `$injector`, you can inject `blueprintService` directly in `saveToCatalogModalDirective`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -64,8 +64,44 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, co
         $scope.buttonText = $scope.config.label || ($scope.config.itemType ? `Update ${$scope.config.name || $scope.config.symbolicName}` : 'Add to catalog');
 
         $scope.activateModal = () => {
-            // Override callback to update catalog configuration data in other applications
-            $scope.config = (composerOverrides.updateCatalogConfig || (($scope, $element) => $scope.config))($scope, $element);
+            function injectorGet(reference) { return $element.injector().get(reference); }
+            let blueprintService = injectorGet('blueprintService');
```
tbouron: Instead of using the `$injector`, you can inject `blueprintService` directly in `saveToCatalogModalDirective`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -64,8 +64,44 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, co
...
+
+            let entity = blueprintService.get();
+            let metadata = blueprintService.entityHasMetadata(entity) ? blueprintService.getEntityMetadata(entity) : { };
+            let config = $scope.$parent.$parent.vm.saveToCatalogConfig;
```
tbouron: You should pass `saveToCatalogConfig` as a parameter for the directive instead of getting it from the parent. You create a link here which is brittle.

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -64,8 +64,44 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, co
         $scope.buttonText = $scope.config.label || ($scope.config.itemType ? `Update ${$scope.config.name || $scope.config.symbolicName}` : 'Add to catalog');
 
         $scope.activateModal = () => {
-            // Override callback to update catalog configuration data in other applications
-            $scope.config = (composerOverrides.updateCatalogConfig || (($scope, $element) => $scope.config))($scope, $element);
+            function injectorGet(reference) { return $element.injector().get(reference); }
+            let blueprintService = injectorGet('blueprintService');
+
+            let entity = blueprintService.get();
+            let metadata = blueprintService.entityHasMetadata(entity) ? blueprintService.getEntityMetadata(entity) : { };
```
tbouron: `metadata` is a `Map`, not an object. So if there isn't any, it should return `new Map()`. It currently throws a JS exception 

Also, why do you get the metadata? There is no need there as the getters for `version`, `icon` and `id` already return this info out of the box.

PR #108, opened 2018-11-13
improve logic and ui for inferring bundle and type name when saving to catalog

ahgittin:
previously showed "untitled" if you clilcked to expand on a new blueprint, now defaults to tidied version of display name (which is required)

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -100,7 +100,7 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, co
             let modalInstance = $uibModal.open({
                 templateUrl: TEMPLATE_MODAL_URL,
                 size: 'save',
-                controller: ['$scope', 'blueprintService', 'paletteApi', 'brUtilsGeneral', CatalogItemModalController],
+                controller: ['$scope', '$filter', 'blueprintService', 'paletteApi', 'brUtilsGeneral', CatalogItemModalController],
                 scope: $scope,
             });
 
@@ -119,7 +119,7 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, co
     }
 }
 
-export function CatalogItemModalController($scope, blueprintService, paletteApi, brUtilsGeneral) {
+export function CatalogItemModalController($scope, $filter, blueprintService, paletteApi, brUtilsGeneral) {
     $scope.REASONS = REASONS;
     $scope.VIEWS = VIEWS;
     $scope.TYPES = TYPES;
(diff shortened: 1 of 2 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -223,3 +228,7 @@ function templateCache($templateCache) {
     $templateCache.put(TEMPLATE_URL, template);
     $templateCache.put(TEMPLATE_MODAL_URL, modalTemplate);
 }
+
+var bundlize = (input) => input && input.split(/[^a-zA-Z0-9]+/).filter(x => x).join('-').toLowerCase();
```
tbouron: Why create this `var` (which should be `let`) ? The function should live within `bundlizeProvider` which you can then call directly within this file with `bundlizeProvider()(input);`. Actually, as better way would be to rely on the `$filter` service like so: `$filter('bundlize`)(input);`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
```diff
@@ -57,19 +60,17 @@ <h3 class="modal-title">{{getTitle()}}</h3>
...
                 <p class="help-block" ng-show="form.bundle.$invalid">
-                    <span ng-if="form.bundle.$error.required">You must specify a bundle ID</span>
                     <span ng-if="form.bundle.$error.pattern">The bundle ID can contains only letters, numbers as well a the following characters: <code>.</code>, <code>-</code> and <code>_</code></span>
                 </p>
             </div>
 
             <div class="form-group" ng-class="{'has-error': form.symbolicName.$invalid}">
                 <label class="control-label">Blueprint symbolic name</label>
-                <input ng-model="config.symbolicName" ng-disabled="state.saving" class="form-control" placeholder="E.g my-catalog-id" name="symbolicName" required ng-pattern="state.pattern" autofocus />
+                <input ng-model="config.symbolicName" ng-disabled="state.saving" class="form-control" name="symbolicName" ng-pattern="state.pattern" autofocus placeholder="{{ config.name | bundlize }}"/>
```
tbouron: I get that `config.name` is required but until you set it, there will be no placeholder. So you should do `{{(config.name | bundlize) || 'E.g my-catalog-id'}}`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
```diff
@@ -57,19 +60,17 @@ <h3 class="modal-title">{{getTitle()}}</h3>
                 <label class="control-label">Bundle ID</label>
                 <div class="input-group">
                     <span class="input-group-addon">catalog-bom-</span>
-                    <input ng-model="config.bundle" ng-disabled="state.saving" class="form-control" placeholder="E.g my-bundle" name="bundle" required ng-pattern="state.pattern" autofocus />
+                    <input ng-model="config.bundle" ng-disabled="state.saving" class="form-control" name="bundle" ng-pattern="state.pattern" autofocus placeholder="{{ config.name | bundlize }}"/>
```
tbouron: I get that `config.name` is required but until you set it, there will be no placeholder. So you should do `{{(config.name | bundlize) || 'E.g my-bundle'}}`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
ahgittin: i didn't like `bundlizeProvider()(textToBundlize)`, but `$filter('bundlize')(text)` is not bad, changed

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
ahgittin: good spot, fitxed

ahgittin:
good comments, all addressed

PR #109, opened 2018-11-16
catalog saver to support 'application', 'template', and 'entity'

ahgittin:
in line with https://github.com/apache/brooklyn-server/pull/1015
also when using a template the metadata is not populated as the intention is probably not to overwrite,
unless the user selects to save it as a template

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -44,7 +44,7 @@ const TYPES = [
 angular.module(MODULE_NAME, [angularAnimate, uibModal, brUtils])
     .directive('catalogSaver', ['$rootScope', '$uibModal', '$injector', '$filter', 'composerOverrides', 'blueprintService', saveToCatalogModalDirective])
     .directive('catalogVersion', ['$parse', catalogVersionDirective])
-    .directive('blueprintNameOrSymbolicNameAndBundleIdRequired', blueprintNameOrSymbolicNameAndBundleIdRequiredDirective)
+    .directive('composerBlueprintNameValidator', composerBlueprintNameValidatorDirective)
     .filter('bundlize', bundlizeProvider)
     .run(['$templateCache', templateCache]);
 
@@ -210,7 +210,7 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
         $scope.defaultSymbolicName = ($scope.config.itemType==='template' && $scope.config.original.symbolicName) || bundlize(newName) || null;
         $scope.defaultBundle = ($scope.config.itemType==='template' && $scope.config.original.bundle) || bundlize(newName) || null;
     };
-    $scope.$watchGroup(['config.name', 'config.itemType'], (newVals) => {
+    $scope.$watchGroup(['config.name', 'config.itemType', 'config.bundle', 'config.symbolicName'], (newVals) => {
         $scope.updateDefaults(newVals[0]);
         $scope.form.name.$validate();
         $scope.buttonText = $scope.buttonTextFn();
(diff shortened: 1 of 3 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -42,14 +42,15 @@ const TYPES = [
 ];
 
 angular.module(MODULE_NAME, [angularAnimate, uibModal, brUtils])
-    .directive('catalogSaver', ['$rootScope', '$uibModal', '$injector', 'composerOverrides', 'blueprintService', saveToCatalogModalDirective])
+    .directive('catalogSaver', ['$rootScope', '$uibModal', '$injector', '$filter', 'composerOverrides', 'blueprintService', saveToCatalogModalDirective])
     .directive('catalogVersion', ['$parse', catalogVersionDirective])
+    .directive('blueprintNameOrSymbolicNameAndBundleIdRequired', blueprintNameOrSymbolicNameAndBundleIdRequiredDirective)
```
tbouron: Wow, that a very loooooong name. Could we come up with something shorter like `blueprintNameValidator`? The name doesn't need to say what it does exactly

ui-modules/blueprint-composer/app/views/main/main.controller.js
```diff
@@ -105,16 +105,31 @@ export function MainController($scope, $element, $log, $state, $stateParams, brB
 
     vm.saveToCatalogConfig = {};
     if (edit) {
+        console.log("edit", edit);
```
tbouron: To remove

ahgittin:
@tbouron agree it would be nice to make the differences clearer but not sure how.  the test to apply is whether it's clearer than before where we prompted for "Application" or "Entity" with no difference, and I think it is.  i've expanded the wording slightly, and welcome further enhancements, but anything more than quick fixes I think will be a longer discussion and shouldn't block this:

* *Application entity*: Save as an application entity which can be deployed on its own, or configured and used in blueprints but only config and sensors declared at the root are accessible in the Composer ('application' item type)
* *Application template*: Save as a blueprint template which can be used as an editable starting point for blueprints or used as an application entity ('template' item type)
* *Extended entity*: Save as an entity which can be configured and used in blueprints, exposing the config and adjuncts it inherits ('entity' item type)

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
ahgittin: okay, compromise, changed to `composerBlueprintNameValidator`

ui-modules/blueprint-composer/app/views/main/main.controller.js
ahgittin: good spot

ahgittin:
one other change, we weren't watching `bundle` and `symbolicName` so the validator didn't do everything it should, that's fixed too now

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
tbouron: Sounds good 👍

aledsage:
@ahgittin for 'template', is that really what it means in the product from a user's perspective? Something declared as a 'template' will appear in the quick launch, whereas other things will not. When you click on it in the 'quick launch', then a dialog pops up with two buttons: `deploy` and `open in editor`. The latter opens a simple text editor in the quick-launch dialog, rather than re-directing to the blueprint composer.

Your description of 'template' is accurate for when this was originally added to the code base, but I don't think that is what it does now, or how it is used now. That description won't make sense to a user, based on how the Brooklyn UI behaves.

tbouron:
@ahgittin I'm wondering if the tooltip texts should include the item type, the same way you describe them in your previous comment. I think it's a good compromise so users have a reference they can scan for.

ahgittin:
@tbouron the itemType is included in the tooltip text, in parentheses at the end

@aledsage i think you're flagging a problem with quick launch that it is cheating a bit using the `catalog_template` tag for what should be a special `quick_launch` or favourites mechanism.  that should be fixed as a separate task.  but until we've done this i'll add to that tooltip text something like that "In some contexts this prioritizes the blueprint for inclusion in quick-selection views."

tbouron:
@aledsage Are you happy to address your comment in a subsequent PR? I'm ok to merge this one but would like to check with you first.

aledsage:
Sure, happy to merge and we can make subsequent improvements in other PRs. Merging now.

PR #111, opened 2018-11-20
fix json editor state issues

ahgittin:
if you are editting json in the spec editor, every valid change does a model update and changes the `$viewValue` to the stringified output `{"key":"vallue"}` which is unhelpful when you're trying to get useful spaces

this prevents the update if the `$viewValue` gives the same value, and if making a new value it nicely formats it


aledsage:
LGTM; merging.

PR #113, opened 2018-11-22
Fix template overrides - decorate `script` directive to prevent re-overriding of templates

ahgittin:
by default `<script>` will install to template cache whenever it is processed as a directive.
this decorates it to be no-op if there is already something in the cache with that name,
allowing programmatic overrides of templates configured using `<script id="...">` notation.

### DIFF
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
@@ -36,7 +36,10 @@ const REPLACED_DSL_ENTITYSPEC = '___brooklyn:entitySpec';
 angular.module(MODULE_NAME, [onEnter, autoGrow, blurOnEnter, brooklynDslEditor, brooklynDslViewer])
     .directive('specEditor', ['$rootScope', '$templateCache', '$injector', '$sanitize', '$filter', '$log', '$sce', '$timeout', '$document', '$state', '$compile', 'blueprintService', 'composerOverrides', specEditorDirective])
     .filter('specEditorConfig', specEditorConfigFilter)
-    .filter('specEditorType', specEditorTypeFilter);
+    .filter('specEditorType', specEditorTypeFilter)
+    .run(['$templateCache', templateCache]);
+
+const TEMPLATE_URL = 'blueprint-composer/component/spec-editor/spec-editor.template.html';
 
 export default MODULE_NAME;
 
@@ -79,7 +82,7 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             model: '='
         },
         controller: ['$scope', '$element', controller],
-        template: template,
+        templateUrl: TEMPLATE_URL,
         link: link,
         controllerAs: 'specEditor',
     };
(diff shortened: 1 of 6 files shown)


sferot:
It works fine for me :+1:

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -79,7 +82,7 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             model: '='
         },
         controller: ['$scope', '$element', controller],
-        template: template,
+        templateUrl: TEMPLATE_URL,
```
tbouron: While do do this, could you change it to match what the other directive do? https://github.com/apache/brooklyn-ui/blob/b04423b169e0d9a749ace79ccd446b4c6f3bb232/ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js#L35-L37

So it can be overridden directly from the attribute `template-url` on the directive itself.

ui-modules/utils/script-tag-non-overwrite/script-tag-non-overwrite.js
```diff
@@ -0,0 +1,48 @@
...
+
+export default MODULE_NAME;
+
+const BROOKLYN_CONFIG = 'brooklyn.config';
+
+function scriptTagDirectiveDecorator($delegate, $templateCache) {
+    let base = $delegate[0];
+    return [ Object.assign({}, base, { compile: function(el, attr) {
+        let match = $templateCache.get(attr.id);
+        if (!(match === null || typeof match === 'undefined')) {
```
tbouron: I would be reluctant to change the default behaviour for the entire application without a possibility to go back. I think it would be wise to add a `override` (or a better name) parameter to revert to the default behaviour is we need/want to

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
ahgittin: good spot, sorry i missed this!

ui-modules/utils/script-tag-non-overwrite/script-tag-non-overwrite.js
ahgittin: i wondered about this too.  we have unique names for all our things and i don't see any way someone would have registered a template that isn't meant to override our things, so didn't think we needed an "opt-out".

however there is a risk with 3rd party libraries that templates might have non-unique names and this could cause an issue there.  doesn't seem to be a problem but if we're concerned what i'd suggest is that we make this behaviour "opt-in", checking an extra attribute e.g. `defer-to-preexisting-id`, which we set everywhere in our code that we define a `<script id="..." ...` tag.

not sure it's needed but will add if you think so @tbouron .

tbouron: @ahgittin Aye, that was the use-case I had in mind. I don't think it's strictly speaking required but better safe than sorry => I think it would be better to do this change.

Although, there is another solution that is even simpler: don't decorate the `script` directive and update the way we register template to do that in a ` .config` block (like the main template) rather than using `<script>` tags

PR #115, opened 2018-12-05
Separate function of left and right panels

sferot:
As proposed in the mailing list we implemented some modifications to the three panels view.
We separated the functionalities : We add every kind of items with the left panel, and all the configuration is done with the right panel.
We also add a way to reduce/close the right panel. It will stay closed until voluntarily reopened.
You can see the result in the following video : https://webmshare.com/ayKMQ


nakomis:
@sferot Could you rebase this against master please, it would be nice to get this merged

PR #117, opened 2018-12-14
require login to webconsole for webapps

ahgittin:
the REST API has always been secured but we now secure these static assets also;
this gives a better experience on login, and it will be even more important when
we introduce oauth-based logins where the redirect page (served by the LoginModule
defined in the REST API module, connected to the LoginService defined in the jetty bundle,
both in brooklyn-server) should be served in response to the request for index.html,
not just on the REST calls

### DIFF
--- a/modularity-server/features/pom.xml
+++ b/modularity-server/features/pom.xml
@@ -63,46 +63,11 @@
             <plugin>
                 <groupId>org.apache.karaf.tooling</groupId>
                 <artifactId>karaf-maven-plugin</artifactId>
+                <version>${karaf.plugin.version}</version>
+                <extensions>true</extensions>
                 <configuration>
                     <startLevel>100</startLevel>
-                    <aggregateFeatures>true</aggregateFeatures>		
-                    <resolver>(obr)</resolver>
                 </configuration>
-                <executions>
-                    <execution>
-                        <id>verify-brooklyn-ui-modularity-feature</id>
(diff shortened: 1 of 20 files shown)


ahgittin:
fixed a bug reported by @frogfather where one snippet had been added twice

but have noticed that while this works for the first launch, something in the first launch creates a rogue `etc/jetty.xml` file and on the _second_ launch it breaks as the `LoginService` is no longer registered :(

[so DO NOT MERGE yet]

tbouron:
@ahgittin This should be ok for basic auth but I think will fail for any other kind of auth (if implemented in the future) All resources are now blocked by your change, meaning that any HTML nor JS can be read until the user login. 

Again, not a issue for basic auth but this will bite back if (or when) we will move to another kind of auth: oAuth for instance will require to redirect to a provider. Maybe this can be done directly at the jetty level but not sure. Similarly, If we handle login via a login page, this will need to be accessible.

ahgittin:
good spot @tbouron -- i was trying to do things the "right way" with a JaaS login module but that seems geared around the `<login-config>` indeed is limited to BASIC auth.  @jcabrerizo @frogfather and i have come to the conclusion to abort this in favour of a security filter.

ahgittin:
Changed to use the `BrooklynSecurityProviderFilter` introduced in https://github.com/apache/brooklyn-server/pull/1024

PR #118, opened 2019-01-23
[WIP] Parameter UI additions

grkvlt:
Adding code to handle the `brooklyn.parameters` section in a blueprint. Supports editing as JSON but still needs work on editing the parameters in their own dialog.

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -68,6 +68,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
         populateEntityFromApi: populateEntityFromApiSuccess,
         populateLocationFromApi: populateLocationFromApiSuccess,
         addConfigKeyDefinition: addConfigKeyDefinition,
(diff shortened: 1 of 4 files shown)


ahgittin:
This will be really useful, defining parameters for use when adding to the catalog.

Early review of functionality:

* "Parameters" only need to be defined on the root node I think, should not be available elsewhere (what would be the use case for that?)
* I guess the non-JSON edit mode is being worked on?
* Once a parameter is added, it should be available in the config section (probably)

Two bigger related questions:

* Should we replace the funnel and plus with a single "search/add" icon (eg a magnifying glass with a + in the glass?) which combines the two modes?  (Both for config and for parameters.)
* Should we also have catalog metadata on the root node, and parameters would be part of that?  They could be useful even when not saving to catalog (to define a value once and use it in multiple places) but that could be done with ad hoc config keys, and for the most part parameters will be set when saving to catalog, and I've not been a fan of that info only appearing when you get a popup dialog.

@tbouron @aledsage thoughts on ^ ?

ahgittin:
Superseded by #129

tbouron:
As @ahgittin, this one is superseded by #129 so closing it
But thanks @grkvlt for starting it 👍

PR #119, opened 2019-01-24
tables should run with their own scope

ahgittin:
without this if you put _two_ tables in the same parent controller view they collide on their data and have their own data

### DIFF
--- a/ui-modules/utils/table/index.js
+++ b/ui-modules/utils/table/index.js
@@ -103,6 +103,7 @@ export function brTableDirective($log) {
         link: link,
         controller: ['$templateCache', 'brUtilsGeneral', controller],
         controllerAs: 'ctrl',
+        scope: true,
         templateUrl: function(element, attrs) {
             return attrs.templateUrl || TEMPLATE_CONTAINER_URL;
         }


tbouron:
I'm really not sure about the implication of this @ahgittin. Some tables have custom templates for columns that use the current scope to get the data from. Creating a child scope for each table will break this pattern

ahgittin:
As discussed, I think you are right, it is a little more subtle (1) if attrs refer to parent scope vars, and (2) if templates refer to parent scope vars.  But we do want enough separation/isolation to be able to put multiple tables in the same view without them clobbering each other.

Will continue investigating.

ahgittin:
the effect of `scope: true` is to create a new scope that prototypically inherits from the parent scope.  i think this is what we want in order to have the table's `scope.ctrl` be unique from another instance of table in the parent.

it can still see all fields from the parent _that were defined when the table was created_.  in worst case, a caller/parent may have to initialize a field in the scope prior to creating the table (eg in the parent's controller) whereas previously they did not (since the scope was shared) and scope fields could be lazilly initialized.

i've tested with complex tables and with a one line fix to initialize a scope field to be a null list, it works fine with `{scope: true}`.

tbouron:
Great, sounds good enough for now @ahgittin, but we should keep an eye on it just in case.

PR #120, opened 2019-01-28
tidy logout and other misc minor changes for new server security model

ahgittin:
update logout to use angular `$http` and actually to call the server to log out

previously it wasn't calling to the server (not sure where it was going wrong but it wasn't working for me!)


ahgittin:
this PR includes a lot of exploration, and i'm _still_ exploring.  it's proving quite hard to get the browser to forget or re-question its cached user:pass for basic auth

ahgittin:
works well alongside https://github.com/apache/brooklyn-server/pull/1032

ui-modules/logout/app/views/main/main.controller.js
```diff
@@ -28,51 +28,123 @@ angular.module(MODULE_NAME, [uiRouter])
...
+    /* There is a lot of complexity in here to support debug pathways with confirmation, 
+     * use of http GET instead of POST, and use of API which returns 200 instead of 401.
+     * This is because logging out nicely is quite tricky.
+     *   Currently we think we have a good pathway without any of that complexity,
+     * so if you haven't set "?debug=true" or other special option in the URL it is
+     * mostly disabled and follows the happy path where it just logs out and prompts
+     * you to log back in. But the debug stuff is left in, in case we encounter edge cases.
+     */
+     
+    $scope.debug = $stateParams.debug;
```
tbouron: You don't need this flag and actually won't work in production as all logging is prevented.
Should be removed then, and use only `$log.debug()`

ui-modules/logout/app/views/main/main.controller.js
```diff
@@ -28,51 +28,123 @@ angular.module(MODULE_NAME, [uiRouter])
...
+    if ($scope.debug) {
+        $log.info("Logout page running in debug mode. state=", $state, "state params=", $stateParams);
+    }
+    if ($stateParams.salt) {
+        // specify some salt to ensure links change in dev mode
+        $scope.salt = (parseInt($stateParams.salt) || 0);
```
tbouron: I don't get what this is for, it is not used anywhere in the controller, only in the template to generate a link. Could you explain how it works?

ui-modules/logout/app/index.js
```diff
@@ -26,10 +26,11 @@ import brCore from 'brooklyn-ui-utils/br-core/br-core';
 import brInterstitialSpinner from 'brooklyn-ui-utils/interstitial-spinner/interstitial-spinner';
 import brooklynModuleLinks from 'brooklyn-ui-utils/module-links/module-links';
 import brooklynUserManagement from 'brooklyn-ui-utils/user-management/user-management';
+import brServerStatus from 'brooklyn-ui-utils/server-status/server-status';
 
 import mainState from './views/main/main.controller';
 
-angular.module('app', [ngAnimate, ngCookies, uiRouter, brCore, brInterstitialSpinner, brooklynModuleLinks, brooklynUserManagement, mainState])
+angular.module('app', [ngAnimate, ngCookies, uiRouter, brCore, brInterstitialSpinner, brServerStatus, brooklynModuleLinks, brooklynUserManagement, mainState])
```
tbouron: Why do we want the `brServerStatus` directive in there? I don't think it useful to have on the logout page

ui-modules/logout/app/views/main/main.controller.js
ahgittin: downstream projects may set this, and we might want to configure info logging in production.  take your point but don't think it's worth changing.

ui-modules/logout/app/views/main/main.controller.js
ahgittin: this makes the links change so that when you click on them they do something.  otherwise the url is the same so nothing happens when running in dev mode (as you are already at the page `/`)

ui-modules/logout/app/index.js
ahgittin: it creates a repeating call to the server on a page in this module so we easily confirm whether it is logged out or not.

PR #121, opened 2019-02-01
Fix the inline edit configuration for the policies in app inspector

tbouron:
This was broken in Brooklyn previously: a white empty box was displayed instead of the form to edit the value.

This changes makes it cleaner and working as expected

fix-policy-config


ahgittin:
does what we want - good fix @tbouron .  merging.

PR #122, opened 2019-02-08
Change NO-CONNECTION error message

jcabrerizo:
NO-CONNECTION message changed for the case something goes wrong with the session after some time of inactivity


ahgittin:
LGTM

PR #124, opened 2019-02-14
DSL Viewer ignored literals; this fixes that and tidies the code a bit

ahgittin:
note that *DSL Editor* still doesn't support literal

### DIFF
--- a/ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
+++ b/ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
@@ -18,18 +18,18 @@
  */
 import angular from 'angular';
 import template from './dsl-viewer.template.html';
-import {KIND} from '../util/model/dsl.model';
+import {KIND, FAMILY} from '../util/model/dsl.model';
 
 const MODULE_NAME = 'brooklyn.components.dsl-viewer';
 const TEMPLATE_URL = 'blueprint-composer/component/dsl-viewer/index.html';
 
 angular.module(MODULE_NAME, [])
-    .directive('dslViewer', dslViewerDirective)
+    .directive('dslViewer', ['$log', dslViewerDirective])
     .run(['$templateCache', templateCache]);
 
 export default MODULE_NAME;
 
-export function dslViewerDirective() {
+export function dslViewerDirective($log) {
     return {
         restrict: 'E',
         templateUrl: function (tElement, tAttrs) {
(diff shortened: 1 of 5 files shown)


ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
```diff
@@ -42,18 +42,24 @@ export function dslViewerDirective() {
     };
 
     function link(scope) {
-        scope.isTargetDsl = (dsl) => {
-            return dsl.kind === KIND.TARGET;
+        
+        scope.getIcon = (dsl) => {
```
tbouron: The template now uses `ng-switch` so there is no point of having this function. The class can be directly put into the HTML element. Less code for same result, yay 🎉

ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.template.html
```diff
@@ -46,6 +48,10 @@
             <dsl-viewer dsl="argument"></dsl-viewer>
         </li>
     </ol>
-</span>
+  </span>
 
-<span ng-if="isLiteralDsl(dsl)">{{dsl.name}}</span>
\ No newline at end of file
+  <span ng-switch-default>
+    {{ getMode(dsl) }}: {{ dsl }}
```
tbouron: the space here is very small, adding the mode is a waste of valuable real estate. We already have an icon to symbolise what type this is, adding the mode is an unnecessary repetition. I think it should be dropped.

You could however add that has a hover title of tooltip

ahgittin: in this default case we do _not_ have an icon so i think this is the right way to do it (but i don't think we have any default cases anymore so it's moot)

ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
ahgittin: I don't follow.  the `ng-switch` is on the mode, but the way the code works (unchanged) is that the icon is based on the name, not the mode.  So I think we still need an icon lookup somewhere.  I think JS is better than doing it in HTML.

ahgittin:
@tbouron thanks for reviewing and flagging these two but for the reasons above I think they are best as is

ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
tbouron: Good point regarding the mode vs name. However, I strongly disagree about doing this in Javascript. This is not the angular way because of performances. Calling the method in `ng-class` will force angular to compute this data on every digest cycle, whereas doing it in the HTML will trigger the evaluation only if the DSL name changes.

ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.template.html
tbouron: Fair enough

ui-modules/blueprint-composer/app/components/dsl-viewer/dsl-viewer.js
ahgittin: good point; now uses `scope.$watch`; also a few other fixes

ahgittin:
significant update @tbouron -- discovered some things broke on nested evaluation, and made some minor tidies.  see last commit for more details.

Untitled

PR #125, opened 2019-02-22
Fix errors around locations in composer

ahgittin:
* now supports map syntax in the visual editor (previously gave errors)
* now works without complaining if no services set
* leaves a marker in place if a location is removed (so consumers can avoid applying defaults or reset defaults)
* fixes bug where images broke during d3 transitions
* rationalises style files, moving similar things together and promoting some shareable things

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -265,15 +265,28 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
         return deferred.promise;
     }
 
+    function locationType(location) {
+        if (!location || typeof location === 'string') return location;
+        if (typeof location === 'object' && Object.keys(location).length==1) return Object.keys(location)[0];
+        return null;
+    }
+    
     function refreshLocationMetadata(entity) {
         let deferred = $q.defer();
 
         if (entity.hasLocation()) {
-            paletteApi.getLocation(entity.location).then((location)=> {
-                deferred.resolve(populateLocationFromApiSuccess(entity, location.catalog || location));
-            }).catch(function () {
-                deferred.resolve(populateLocationFromApiError(entity));
-            });
+            let type = locationType(entity.location);
+            if (type.startsWith("jclouds:")) {
+                // types eg jclouds:aws-ec2 are low-level, not in the catalog
+                deferred.resolve(populateLocationFromApiSuccess(entity, { yamlHere: entity.location }));
+            } else {
+                paletteApi.getLocation(locationType(entity.location)).then((location)=> {
+                    let loc = Object.assign({}, location.catalog || location, { yamlHere: entity.location });
+                    deferred.resolve(populateLocationFromApiSuccess(entity, loc));
+                }).catch(function () {
+                    deferred.resolve(populateLocationFromApiError(entity));
+                });
+            }
         } else {
             deferred.resolve(entity);
         }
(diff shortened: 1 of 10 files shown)


ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -603,16 +616,24 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
...
+        
+        // use icon on item, but if none then generate using *yaml* to distinguish when someone has changed it
+        // (especially for things like jclouds:aws-ec2 -- the config is more interesting than the type name)
+        entity.miscData.set('locationIcon', data==null ? null : data.iconUrl || iconGenerator(data.yamlHere ? JSON.stringify(data.yamlHere) : data.symbolicName));
         return entity;
     }
 
     function populateLocationFromApiError(entity) {
-        entity.clearIssues({group: 'location'});
+        populateLocationFromApiSuccess(entity, { yamlHere: entity.location });
```
tbouron: That's wrong. If the location can be setup from the entity, then this code should be put into the caller function, not here. This function is to setup the state if there is an error when setting up the location.

Also, you want to keep the `entity.clearIssues({group: 'location'});` because errors are recreated just below. Not doing so will result to the same error displayed twice

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -603,16 +616,24 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService)
 
     function populateLocationFromApiSuccess(entity, data) {
         entity.clearIssues({group: 'location'});
-        entity.location = data.symbolicName;
-        entity.miscData.set('locationName', data.name);
-        entity.miscData.set('locationIcon', data.iconUrl || iconGenerator(data.symbolicName));
+        entity.location = data.yamlHere || data.symbolicName;
```
tbouron: `yamlHere` property feels odd. I would go for just `yaml` or `spec`

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -254,6 +254,9 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
         }
         scope.nonempty = (o) => o && Object.keys(o).length;
         scope.defined = specEditor.defined = (o) => (typeof o !== 'undefined');
+        specEditor.isInstance = (x, type) => (typeof x === type);
+        specEditor.json = x => JSON.stringify(x, null, 2);
```
tbouron: No need for this, there is a pipe called `json` that you can use like this `{{ data | json }}`

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
```diff
@@ -362,7 +362,12 @@ <h4>No location attached</h4>
             <p ng-repeat="issue in state.issues | filter:{group:'location'}" class="alert alert-{{issue.severity}}">
                 <em ng-bind-html="issue.message"></em>
             </p>
-            <p>Will be deployed to: <strong>{{model.miscData.get('locationName')}}</strong></p>
+            <p>
+                Targeted at: 
+                <strong ng-if="specEditor.isInstance(model.location, 'string')">{{ model.miscData.get('locationName') }}</strong>
+                <pre ng-if="!specEditor.isInstance(model.location, 'string')">{{ specEditor.json(model.location) }}</pre>
```
tbouron: See my comment above

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
```diff
@@ -281,6 +282,13 @@ export class Entity {
      */
     removeLocation() {
         LOCATIONS.delete(this);
+        this.miscData.delete('locationName');
+        this.miscData.delete('locationIcon');
+        
+        // this field provides a way for consumers to detect if the location was explicitly removed;
+        // this can be useful to prevent default locations from being applied
+        this.miscData.set('locationRemoved', true);
```
tbouron: I don't get what this is for, even with the comment. Can you explain the use-case please?

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
ahgittin: i wanted something that obviously wouldn't conflict.  it could be `yamlOverride` but to me, `Here` conveys that it is _not_ part of any external spec.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
ahgittin: in a downstream project we automatically add a location if none is present.  however if the user removes it in the UI, we in that downstream code want to now that we are able _not_ to re-add it in that case.

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
ahgittin: from a UX point of view, if there is an error we should indicate this fact (which this method now does) but it should still make best-effort to reflect what is known about the location in the UI (which the call to `populateOnSuccess` does), in addition to that error.  the call to `populateOnSuccess` clears the issues (so we don't get double), it also sets the ` `locationName` sensibly, so by calling that now (the changes I make here) we avoid code duplication.  i think the behaviour as changed is best.

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
ahgittin: nice, i'll revert this

ahgittin:
fixed the json, i think the rest can stand

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
tbouron: Hum, I get your point. I think I'm complaining mainly because the name is not right, this should be `populateLocation` instead of `populateLocationFromApiSuccess`. Could you change that?

ahgittin: :) yes, i concur!

ahgittin:
all addressed

PR #127, opened 2019-03-11
Allow bigger zoom in, for large screen mainly

tbouron:
This is useful when working on a large screen. We can now zoom in 30% more

### DIFF
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -173,7 +173,7 @@ export function D3Blueprint(container) {
         relationships: [],
     };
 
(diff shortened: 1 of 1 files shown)


ahgittin:
would it be worth making the scale extent some step function of display size, as on normal screens the old max is right, and on a hypothetical 10k display 1.3 isn't much?  something like (untested):

`maxExtent = Math.max(1, Window.width/1600)`

tbouron:
@ahgittin Good idea, but the function looks more like `log`, i.e for:
- < 2k => `1`
- 2k => `1`
- 4k => `1.3`
- 8k => `1.6`
- etc

So I would suggest the following function: `Math.max(1, 1+log(windowWidth/1024))` where `windowWidth/1024` should return `2` for 2k resolution and so on.

I just pushed the changes

ahgittin:
+1 to an apposite use of logarithms!  Needs to be `Math.log` not plain `log` but otherwise great, amended and merged.

PR #129, opened 2019-03-26
UI for defining parameters

ahgittin:
has a field-based mode and JSON mode

builds on #118 , makes it mergeable

ideally flaws in the PR can be tracked here but addressed in follow-on PRs unless critical.  i'm aware of the following wishlist items:

1. if no params, it's not obvious how to add ("No params" should have option to "Add" one)
2. are the icons in top-right of params section needed (filter etc)
3. if same parameter entered twice you get angular errors
4. parameters section should appear only on application (it's pointless elsewhere)
5. parameters added should show up in the config section (maybe?)
6. do all the fields work as expected when defining a parameter used subsequently (eg label, constraint)
7. parameters are mostly pointless if you're deploying directly, it's mainly for saving and reuse (or some types of export)

i think 1,4,6 need addressed imminently but i don't think they should block merging this

### DIFF
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
@@ -83,6 +83,11 @@ export const CONFIG_FILTERS = [
     }
 ];
 
+export const PARAM_TYPES = [
+    // cut-down list from BasicSpecParameter.ParseYamlInputs
+    'string', 'boolean', 'integer', 'double', 'duration', ' port'
+    ];
+
 export function specEditorDirective($rootScope, $templateCache, $injector, $sanitize, $filter, $log, $sce, $timeout, $document, $state, $compile, blueprintService, composerOverrides) {
     return {
         restrict: 'E',
@@ -112,6 +117,7 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
         scope.REPLACED_DSL_ENTITYSPEC = REPLACED_DSL_ENTITYSPEC;
         scope.parameters = [];
         scope.config = {};
+        specEditor.paramTypes = PARAM_TYPES;
 
         specEditor.getParameter = getParameter;
         specEditor.addParameter = addParameter;
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
@@ -47,6 +47,8 @@ export const CONFIG_FILTERS = [
     {
         id: 'suggested',
         label: 'Suggested',
+        icon: 'plus-circle',
+        hoverText: 'Show config keys that marked as pinned or priority',
         filter: (item)=> {
             return item.pinned && item.priority > -1;
         }
@@ -54,6 +56,8 @@ export const CONFIG_FILTERS = [
     {
         id: 'required',
         label: 'Required',
+        icon: 'times-circle',
+        hoverText: 'Show config keys that are required or have an issue',
         filter: (item, model)=> {
             return (item.constraints && item.constraints.required) ||
                 (model && model.issues && model.issues.some((issue)=>(issue.group === 'config' && issue.ref === item.name)) );
@@ -62,6 +66,8 @@ export const CONFIG_FILTERS = [
     {
         id: 'inuse',
         label: 'In Use',
+        icon: 'check-circle',
+        hoverText: 'Show config keys that are in use',
         filter: (item, model)=> {
             return model && model.config && model.config.has(item.name);
         }
@@ -69,6 +75,8 @@ export const CONFIG_FILTERS = [
     {
         id: 'all',
         label: 'All',
+        icon: 'th-list',
+        hoverText: 'Show all config keys',
         filter: (item)=> {
             return item;
         }
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
@@ -99,7 +99,6 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
 
     function link(scope, element, attrs, specEditor) {
         scope.specEditor = specEditor;
-        scope.addConfigKey = addConfigKey;
         scope.FAMILIES = EntityFamily;
         scope.RESERVED_KEYS = RESERVED_KEYS;
         scope.REPLACED_DSL_ENTITYSPEC = REPLACED_DSL_ENTITYSPEC;
@@ -783,6 +782,11 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
                 return config;
             });
         }
+        specEditor.getConfig = getConfig;
+        function getConfig(name) {
+            return scope.model.miscData.get('config').find(p => p.name === name);
+        }
+        
 
         function loadCustomConfigWidgetMetadata(model) {
             let customConfigWidgets = (scope.model.miscData.get('ui-composer-hints') || {})['config-widgets'] || [];
@@ -995,8 +999,8 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             scope.model.setConfigFromJson(result);
         }
 
-        function addConfigKey() {
-            let name = scope.state.config.add.value;
+        specEditor.addConfigKey = addConfigKey;
+        function addConfigKey(name) {
             if (name) {
                 let allConfig = scope.model.miscData.get('config');
                 blueprintService.addConfigKeyDefinition(allConfig, name);
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
@@ -205,6 +205,10 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             blueprintService.refreshAllRelationships();
         });
 
+        scope.$watch('state.parameters', (newVal, oldVal) => {
+            setModelFromLocalParameters();
+        }, true);
+
         // Parameters
         scope.$watch('parameters', (newVal, oldVal) => {
             setModelFromLocalParameters();
@@ -375,6 +379,12 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
         };
         scope.recordParameterFocus = specEditor.recordParameterFocus = ($item) => {
             scope.state.parameters.focus = $item.name;
+            scope.state.parameters.edit = {
+                item: $item,
+                newName: $item.name,
+                constraints: $item.constraints ? JSON.stringify($item.constraints) : '',
+                json: JSON.stringify($item, null, "  "),
+            };
         };
 
         scope.removeAdjunct = ($event, adjunct) => {
@@ -634,13 +644,11 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             }
         };
         scope.parameterCodeModeClick = (item) => {
-            let oldMode = !!(scope.state.parameters.codeModeActive[item.name]);
+            let oldMode = !!(item && scope.state.parameters.codeModeActive[item.name]);
             let value = null;
             if (oldMode) {
                 // leaving code mode
-                value = scope.state.parameters.edit.value;
-                scope.state.parameters.edit.name = '';
-                scope.state.parameters.edit.open = false;
+                value = scope.state.parameters.edit.json;
                 try {
                     value = JSON.parse(value);
                     let i = scope.parameters.findIndex(p => p.name === item.name);
@@ -652,13 +660,11 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             } else {
                 // entering code mode
                 // leave code mode for other parameter if set
-                if (scope.state.parameters.edit.name !== '') {
-                    scope.parameterCodeModeClick(getParameter(scope.state.parameters.edit.name));
+                if (scope.state.parameters.edit.newName && scope.state.parameters.edit.newName!=item.name) {
+                    scope.parameterCodeModeClick(getParameter(scope.state.parameters.edit.newName));
                 }
                 // convert local parameter from json to non-json or vice-versa
                 value = item;
-                scope.state.parameters.edit.name = item.name;
-                scope.state.parameters.edit.open = true;
                 if (value != null) {
                     try {
                         JSON.parse(value);
@@ -668,7 +674,7 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
                         value = JSON.stringify(value, null, "  ");
                     }
                     if (value != null) {
-                        scope.state.parameters.edit.value = value;
+                        scope.state.parameters.edit.json = value;
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


ahgittin:
i've addressed 1,2,4, and applied a similar toolbar cleanup to configuration (also use labels a la palette instead of checkboxes for the filters).  3 is minimised due to better duplicate prevention logic.

needs some testing for 5 and 6 (but shouldn't block this merge)

image

ahgittin:
did some testing and it all works (6 is happy), except a default value of `''` is being set for things so `required` has no effect.

i've fixed that such that we drop fields if empty (including constraints) except for default value we're slightly more clever if `""` is set in yaml; see last commit for details.

this leaves two things to consider:

* 5 -- i can't think of a use case for including params in the config section, so while logically it is valid to include them, practically i see no reason to, it just makes noise and confusion; setting a `default` is clearer than setting a value and they achieve the same things
* 7 -- not one for now but as mentioned in #118 we might want to merge some of the catalog metadata into the "Application" widget instead of the modal "Save" popup, with an option there "Save this as a catalog item" and parameters only appears if that is checked

jcabrerizo:
Great job you both! I'd like add some ideas:

1. When we add an efector to an application, it shows an small icon on the top right. Have something similar to notice the user that this app has properties woudl help him
image
2. Parameter type is a text input box. It fails when we tray to add something out of [string, bool, boolean, byte, char, character, short, integer, int, long, float, double, duration, timestamp, port], a select input seem to be the correct here
3. Same to the above to Constraints
4. Even names are self-explanatory, I'd add an icon with info to, types, and default value to inform it accepts an empty string

None of this is a blocker so we can do it in future

ahgittin:
all good suggestions @jcabrerizo , TY.  your 2 and 4 are easy, i'll do that in this PR.

not sure about 1, it might make the display noisy.  but we should consider as part of a wider UI review at some point.

i'd like to do a lot more around constraints but there aren't quick wins, as it needs a list and some of the constraints themselves take parameters (eg `requiredUnless("otherKey")`).

ahgittin:
@jcabrerizo both updated; went with a subset of common parameter types, with note in code and ability to write others

good to merge in your opinion?

jcabrerizo:
Yes, looks good to me now except for one weird thing: you can not use `name` as key name. Makes sense?

ahgittin:
that's weird, `name` seems to make it blow up.  i'll look at that.

ahgittin:
took me far too long to find, but i've got it now @jcabrerizo .  can you look at last commit and then we can merge!

jcabrerizo:
Perfect now 👍

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -449,13 +527,25 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
+            return item.widgetMode;
+        };
+        specEditor.getParameterWidgetMode = (item) => {
+            let type = item.type || item.typeName;
+
+            if (type === 'java.lang.Boolean') type = 'boolean';
+            else if (type === 'java.util.Map') type = 'map';
+            else if (type === 'java.util.Set' || type === 'java.util.List' || type === 'java.util.Collection' || type.startsWith('List<')) type = 'array';
+
+            if (type.startsWith('AWS::')) type = 'unknown';
```
tbouron: Why is there a mention of AWS here?

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -503,7 +593,7 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             if (!specEditor.defined(value)) {
                 value = null;
             }
-            if (value!=null) {
+            if (value != null) {
```
tbouron:
```suggestion
            if (value !== null) {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -548,24 +638,65 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
                         value = JSON.stringify(value);
                     }
                 }
-                if (value!=null) {
+                if (value != null) {
```
tbouron: Should use `!==`
```suggestion
                if (value !== null) {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -548,24 +638,65 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
                     scope.config[item.name] = value;
                 }
             }
             scope.state.config.codeModeActive[item.name] = !oldMode;
-            if (value!=null) {
+            if (value != null) {
```
tbouron:
```suggestion
            if (value !== null) {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -548,24 +638,65 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
             } else {
-                return "Edit in simple mode, unwrapping JSON if possible ["+itemName+"]";
+                // entering code mode
+                // leave code mode for other parameter if set
+                if (scope.state.parameters.edit.newName && scope.state.parameters.edit.newName!=item.name) {
+                    scope.parameterCodeModeClick(getParameter(scope.state.parameters.edit.newName));
+                }
+                // convert local parameter from json to non-json or vice-versa
+                value = item;
+                if (value != null) {
```
tbouron:
```suggestion
                if (value !== null) {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -601,36 +738,37 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
+                    $templateCache.put(templateName, '<i>Widget ' + widgetName + ' missing</i>');
                 }
             }
             return templateName;
         };
-
         specEditor.isDsl = (key, index) => {
             let val = scope.model.config.get(key);
-            if (specEditor.defined(val) && specEditor.defined(index) && index!=null) val = val[index];
+            if (specEditor.defined(val) && specEditor.defined(index) && index != null) val = val[index];
```
tbouron:
```suggestion
            if (specEditor.defined(val) && specEditor.defined(index) && index !== null) val = val[index];
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -639,9 +777,9 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
         };
         specEditor.isDslWizardButtonAllowed = (key, index, nonModelValue) => {
             let val = scope.model.config.get(key);
-            if (specEditor.defined(val) && specEditor.defined(index) && index!=null) val = val[index];
-            if (!specEditor.defined(val) || val===null || val==='') val = nonModelValue;
-            if (!specEditor.defined(val) || val===null || val==='') return true;
+            if (specEditor.defined(val) && specEditor.defined(index) && index != null) val = val[index];
```
tbouron:
```suggestion
            if (specEditor.defined(val) && specEditor.defined(index) && index !== null) val = val[index];
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -654,48 +792,61 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
                 }
-
                 result[key] = getLocalConfigValueFromModelValue(key, value);
             }
             scope.config = result;
         }
 
         function getLocalConfigValueFromModelValue(key, value) {
-            if (!specEditor.defined(value) || value==null) {
+            if (!specEditor.defined(value) || value == null) {
```
tbouron:
```suggestion
            if (!specEditor.defined(value) || value === null) {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -788,11 +938,11 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
             }
 
             // all other primitives treat as string (as they will get a string-based widget)
-            return ""+value;
+            return "" + value;
         }
 
         function getModelValueFromString(val) {
-            if (!specEditor.defined(val) || val==null || typeof val !== 'string') {
+            if (!specEditor.defined(val) || val == null || typeof val !== 'string') {
```
tbouron:
```suggestion
            if (!specEditor.defined(val) || val === null || typeof val !== 'string') {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -888,13 +1037,151 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
+                        } else {
+                            scope.state.parameters.edit.errors.push({ message: "Constraint JSON must be a list" });
+                        }
+                    } catch (e) {
+                        // $log.warn("ERROR parsing constraints", scope.state.parameters.edit.constraints, e);
+                        scope.state.parameters.edit.errors.push({ message: "Invalid constraint JSON" });
+                    }
+                    
+                    // empty values are removed
+                    if (item.description == '') {
```
tbouron:
```suggestion
                    if (item.description === '') {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -888,13 +1037,151 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
+                        delete item['description'];
+                    } 
+                    if (item.label == '') {
```
tbouron:
```suggestion
                    if (item.label === '') {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
```diff
@@ -888,13 +1037,151 @@ export function specEditorDirective($rootScope, $templateCache, $injector, $sani
...
+                        delete item['label'];
+                    } 
+                    if (item.default == '') {
```
tbouron:
```suggestion
                    if (item.default === '') {
```

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.directive.js
ahgittin: quite right, this shouldn't be here!

ahgittin:
great points @tbouron 

> `==` vs `===` and `!=` vs `!==` ...

you are 100% correct we should use the latter.  i revert to bad other language habits.  TY.

> (3) is absolutely critical if you have duplicate key. We can live with awkward behaviour but not breaking the UI. This can be solved by adding validation on the key field, to check if any exists already.

yeah, i buy that.  i don't see an easy way to do it with validation for the case where duplicate keys are set in yaml.  (we already did it if you tried to supply duplicate keys in the ui so i think from yaml is the only way.)  what i've done instead is that on model load if we find a duplicate we append a number to it, so `brooklyn.parameters: [ { name: y, type: string }, { name: y, type: string }, { name: y, type: string } ]` becomes `y` `y2` `y3`, with warnings logged to the console.  it's awkward and surprising but at least doesn't break the ui.  we could have yaml model validation errors as well for this though the problem would still manifest in the graphical composer if invalid yaml was passed in so we need a way to handle it here.  the other option would be to refuse to show parameters.

> any constraints added into the the constraint field don't work.  Worse, the value is cleared as soon as I unfocus the field which is very irritating if the constraint is long. And setting the constraints in either JSON mode or directly in YAML will silently remove them.

if it's valid JSON it works, but it must be e.g. `["required"]` -- -have changed message to note it needs the brackets _and_ double quotes.  could improve by supporting yaml parsing or a dedicated editor but going for minimal working capability here.  also agree clearing on unfocus is awkward but of course there's a lot of work to store invalid models if we did something else.  we could maybe forbid the unfocus.

> the change to have a single filter/add field is moot. It's awkward to use and hard to understand what it's doing. This would be high priority to fix IMO but in the sake of getting this PR merged (and the spirit of my first point) it can be addressed in a subsequent PR.

doesn't moot mean "no longer relevant or not worth discussion"?  i think you mean it's a step backwards.  personally i thought having two was awkward and one is fairly self-explanatory.  happy to see further evolution and testing but agree it's such a minor point as to be _moot_ :).

> re the new control for filter/add config and parameter, I think the icons you added are unnecessary in this case: it takes valuable real estate and don't give meaning. For example, the cross for required will be interpreted as "close" or "remove" the required params. I think i would be best to not include them for now, until we find a better way.

i removed checkboxes so in total we have a real estate savings.  i think icons make it look more polished, but agree the `x` is confusing.  have changed it to a `stop-circle` (bit white square in black circle).  again happy to see further evolution but i'd like to keep the icons.  if you feel strongly let's get a tie-break opinion!

ahgittin:
(all comments addressed and pushed @tbouron )

tbouron:
>yeah, i buy that. i don't see an easy way to do it with validation for the case where duplicate keys are set in yaml. (we already did it if you tried to supply duplicate keys in the ui so i think from yaml is the only way.) what i've done instead is that on model load if we find a duplicate we append a number to it, so brooklyn.parameters: [ { name: y, type: string }, { name: y, type: string }, { name: y, type: string } ] becomes y y2 y3, with warnings logged to the console. it's awkward and surprising but at least doesn't break the ui. we could have yaml model validation errors as well for this though the problem would still manifest in the graphical composer if invalid yaml was passed in so we need a way to handle it here. the other option would be to refuse to show parameters.

Sounds good for now 👍 

> if it's valid JSON it works, but it must be e.g. ["required"] -- -have changed message to note it needs the brackets and double quotes. could improve by supporting yaml parsing or a dedicated editor but going for minimal working capability here. also agree clearing on unfocus is awkward but of course there's a lot of work to store invalid models if we did something else. we could maybe forbid the unfocus.

Aaaah it requires JSON! Great for the error message, **it would be also great to specify this in the information popup for the constraints @ahgittin.**

> doesn't moot mean "no longer relevant or not worth discussion"? i think you mean it's a step backwards. personally i thought having two was awkward and one is fairly self-explanatory. happy to see further evolution and testing but agree it's such a minor point as to be moot :).

Probably my flawed French translation: `moot` means `subject to debate, arguable` to me. 

> i removed checkboxes so in total we have a real estate savings. i think icons make it look more polished, but agree the x is confusing. have changed it to a stop-circle (bit white square in black circle). again happy to see further evolution but i'd like to keep the icons. if you feel strongly let's get a tie-break opinion!

Great!

Happy to merge once we have the explanation for the JSON constraints @ahgittin !

ahgittin:
@tbouron the constraints hover text currently says:

> Constraints such as "required" or "forbiddenUnless(otherKey)", expressed as a JSON list (wrapped in square brackets and double quotes)

is this enough?  or did you want more explanation did you have in mind?

would be if we made it html and could use code formatting, we could say

> Constraints such as `[ "required" ]` or `[ "forbiddenUnless(otherKeyNeededIfAndOnlyIfThisOne)", "requiredUnless(otherKeyNeededIfAndOnlyIfThisOne)" ]`, expressed as a JSON list (wrapped in square brackets and double quotes)

but it's fiddly to make it consistent with the others while supporting html so i'm inclined to leave it as is

tbouron:
> > Constraints such as "required" or "forbiddenUnless(otherKey)", expressed as a JSON list (wrapped in square brackets and double quotes)

> is this enough? or did you want more explanation did you have in mind?

Yes, should be enough, happy to merge

PR #130, opened 2019-03-29
Fix entitlement messages

jcabrerizo:
Now shows the messages from the entitlement control system

### DIFF
--- a/ui-modules/app-inspector/app/views/main/inspect/summary/summary.controller.js
+++ b/ui-modules/app-inspector/app/views/main/inspect/summary/summary.controller.js
@@ -188,7 +188,11 @@ export function summaryController($scope, $state, $stateParams, $q, $http, brSna
             entityApi.updateEntityName(applicationId, entityId, this.name).then((response)=> {
                 vm.entity.name = vm.name;
             }).catch((error)=> {
-                brSnackbar.create('Cannot update entity name: the entity [' + entityId + '] is undefined');
+                if(error.data && error.data.error === 403){
+                    brSnackbar.create('Cannot update entity name: ' + error.data.message);
+                }else{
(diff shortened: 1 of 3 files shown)


ui-modules/app-inspector/app/views/main/inspect/summary/summary.controller.js
```diff
@@ -188,7 +188,11 @@ export function summaryController($scope, $state, $stateParams, $q, $http, brSna
             entityApi.updateEntityName(applicationId, entityId, this.name).then((response)=> {
                 vm.entity.name = vm.name;
             }).catch((error)=> {
-                brSnackbar.create('Cannot update entity name: the entity [' + entityId + '] is undefined');
+                if(error.data && error.data.error === 403){
```
tbouron: I don't think you need this change as 403 errors are now handled by the Java securityproviser you did @jcabrerizo ;)

jcabrerizo: The security provider manages the authentication, but an authenticated user can try execute actions that need an entitlement that s/he hasn't. In this case, the exception thrown send an 403 as status. I tried to left the previous behavior and also manage the unauthorized action. If it doesn't make sense for you @tbouron i can change it.

tbouron: Good point about the REST API. But using your argument, it should then be applied to **all endpoints**, not just this one. I think it can be done globally though, like an interceptor of some kind, but it is outside on the scope of this PR IMO.

So I would remove it for now, but create a JIRA ticket to track this task. How does it sound @jcabrerizo ?

jcabrerizo: I've changed the get the message from the `error` object

PR #131, opened 2019-04-22
Avoiding swagger validation

jcabrerizo:
In internet isolated environments is impossible reaching the `swagger.json` file, and an error label is shown in the REST API Doc page.
image

Passing `null` as validator url, the module don't show the validation label as it does when is executed in localhost.


tbouron:
See [windows.ononline](https://developer.mozilla.org/en-US/docs/Web/API/Document/ononline)

ahgittin:
It was my suggestion just to disable validation on the basis that the spec is produced server-side and known to be valid so is there any point in validating it each time?  I don't think there is any degradation in functionality with this -- is there @jcabrerizo ?

tbouron:
Fair point @ahgittin. I'm happy to merge as in then

PR #136, opened 2019-08-22
more tips for imagemin

ahgittin:

PR #137, opened 2019-08-22
more info on imagemin build errors

ahgittin:
replaces #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ You will need the following binaries installed first:
   * `libpng-dev` for Debian/Ubuntu (also requires `libpng12`)
   * `libpng` for MacOS
 
-_Optional, only if the `libpng` cannot be found_
+_Optional, if the `libpng` cannot be found when building (package names may vary; often these are not required):_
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +102,18 @@ cd /path/to/brooklyn_ui
...
 ```
 
+### Errors with libpng
+
+On many systems the `imagemin` dependency of JS shows errors during install, often due to `libpng` missing
+or the `jpeg` compilation failing (note this one can be a non-fatal error at the time, but the build will then fail).
+
+Consult the list above of dependencies to install, and make sure you wipe the `node_modules` dirs.
+
+If it still fails, try installing `nvm` (curl-to-bash`) then `nvm ls-remote` and `nvm ls-install v12.9.0` (or other latest).
```
infrastation: `nvm install v12.9.0`

ahgittin:
This is fixed in my branch but this won't update either, same problem as #136.

I'll leave it for a bit to see if it resolves itself, and if not I'll push the final change manually.

PR #139, opened 2019-09-12
Fix UI modules logs

tbouron:
This does the following:
- enable debug logs for non-production builds
- enable angular runtime logs for non-production builds
- remove the `drop_console` option from webpack to display info, warn and error logs on production builds

### DIFF
--- a/ui-modules/catalog/app/index.js
+++ b/ui-modules/catalog/app/index.js
@@ -45,7 +45,7 @@ angular.module('app', [ngAnimate, ngCookies, ngResource, brCore, brServerStatus,
 
 
 function applicationConfig($logProvider, $compileProvider) {
-    $logProvider.debugEnabled(IS_PRODUCTION);
+    $logProvider.debugEnabled(!IS_PRODUCTION);
     $compileProvider.debugInfoEnabled(!IS_PRODUCTION);
 }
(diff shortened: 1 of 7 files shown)


tbouron:
retest this please

tbouron:
FYI, I change the maven version to be 3.5.4 (latest available on the Apache Jenkins nodes as far as I'm aware)

tbouron:
@geomacy Build pass :shipit:

ui-modules/catalog/app/index.js
```diff
@@ -45,7 +45,7 @@ angular.module('app', [ngAnimate, ngCookies, ngResource, brCore, brServerStatus,
 
 
 function applicationConfig($logProvider, $compileProvider) {
-    $logProvider.debugEnabled(IS_PRODUCTION);
```
ahgittin: `!`  :) good fix

ahgittin:
Thanks guys.  Code looks good, and works as advertised.  I can now use `$log` in my code!

Merging.

geomacy:
> FYI, I change the maven version to be 3.5.4 (latest available on the Apache Jenkins nodes as far as I'm aware)

should have thought of that myself...

PR #140, opened 2019-09-16
Icons on app inspector

ahgittin:
show icons on app-inspector tree view and summary, to help navigate/understand

also optimize/tidy icon generation code

### DIFF
--- a/ui-modules/app-inspector/app/components/entity-tree/entity-node.html
+++ b/ui-modules/app-inspector/app/components/entity-tree/entity-node.html
@@ -22,9 +22,12 @@
             <brooklyn-status-icon value="{{entity.serviceState}}" ng-if="entity.serviceState || entity.applicationId"></brooklyn-status-icon>
             <i class="fa fa-2x fa-external-link" ng-if="!entity.serviceState && !entity.applicationId"></i>
             <span>{{entity.name}}</span>
+            <span class="node-icon"><img ng-src="{{ entity | iconGeneratorPipe:{doNotAutogenerate: true} }}"/></span>
         </a>
-        <div class="entity-node-toggle" ng-if="entity.children.length > 0 || entity.members.length > 0" ng-click="onToggle($event)" >
+        <div class="entity-node-toggle-wrapper">
+          <div class="entity-node-toggle" ng-if="entity.children.length > 0 || entity.members.length > 0" ng-click="onToggle($event)" >
             <span class="glyphicon" ng-class="isChildrenOpen ? 'glyphicon-chevron-up' : 'glyphicon-chevron-down'"></span>
+          </div>
         </div>
     </div>
     <div class="entity-node-children" ng-show="isChildrenOpen">
--- a/ui-modules/app-inspector/app/components/entity-tree/entity-node.html
+++ b/ui-modules/app-inspector/app/components/entity-tree/entity-node.html
@@ -22,7 +22,7 @@
             <brooklyn-status-icon value="{{entity.serviceState}}" ng-if="entity.serviceState || entity.applicationId"></brooklyn-status-icon>
             <i class="fa fa-2x fa-external-link" ng-if="!entity.serviceState && !entity.applicationId"></i>
             <span>{{entity.name}}</span>
-            <span class="node-icon"><img ng-src="{{ entity | iconGeneratorPipe:{doNotAutogenerate: true} }}"/></span>
+            <span class="node-icon"><img ng-src="{{ iconUrl }}"/></span>
         </a>
         <div class="entity-node-toggle-wrapper">
           <div class="entity-node-toggle" ng-if="entity.children.length > 0 || entity.members.length > 0" ng-click="onToggle($event)" >
--- a/ui-modules/app-inspector/app/components/entity-tree/entity-node.less
+++ b/ui-modules/app-inspector/app/components/entity-tree/entity-node.less
@@ -86,7 +86,6 @@
     }
   }
   .node-icon {
-    .pull-right();
     flex-grow: 0 !important;
     max-height: 20px; //let it extend above and below span's boundaries
     overflow: visible !important;
--- a/ui-modules/app-inspector/app/components/entity-tree/entity-node.less
+++ b/ui-modules/app-inspector/app/components/entity-tree/entity-node.less
@@ -75,10 +75,28 @@
         margin-left: 10px;
       }
     }
+    .entity-node-toggle-wrapper {
+      width: 38px;      
+      flex-grow: 0;
+    }
     .entity-node-toggle {
-      flex-shrink: 1;
+      width: 38px;      
       padding: 12px;
       cursor: row-resize;
+    }
+  }
+  .node-icon {
+    .pull-right();
+    flex-grow: 0 !important;
+    max-height: 20px; //let it extend above and below span's boundaries
+    overflow: visible !important;
+    margin-top: -8px;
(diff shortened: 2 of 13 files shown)


ui-modules/app-inspector/app/components/entity-tree/entity-node.less
```diff
@@ -75,10 +75,28 @@
...
+    }
     .entity-node-toggle {
-      flex-shrink: 1;
+      width: 38px;      
       padding: 12px;
       cursor: row-resize;
     }
   }
+  .node-icon {
+    .pull-right();
```
tbouron: Why using `.pull-right()` mixin where the whole block is using flexbox? This is not right, should be one or the other. Based on the current setup, I think this should use flexbox only)

ui-modules/app-inspector/app/components/entity-tree/entity-node.html
```diff
@@ -22,9 +22,12 @@
             <brooklyn-status-icon value="{{entity.serviceState}}" ng-if="entity.serviceState || entity.applicationId"></brooklyn-status-icon>
             <i class="fa fa-2x fa-external-link" ng-if="!entity.serviceState && !entity.applicationId"></i>
             <span>{{entity.name}}</span>
+            <span class="node-icon"><img ng-src="{{ iconUrl }}"/></span>
         </a>
-        <div class="entity-node-toggle" ng-if="entity.children.length > 0 || entity.members.length > 0" ng-click="onToggle($event)" >
+        <div class="entity-node-toggle-wrapper">
```
tbouron: Why wrapping this into another `div`? This entire "row" (i.e. entity status, entity name, icon and expand toggle) can be handled with flexbox which would be cleaner and easier to implement, rather than doing it with unnecessary HTML layout.

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
```diff
@@ -101,7 +97,7 @@ export function entityTreeDirective() {
     }
 }
 
-export function entityNodeDirective() {
+export function entityNodeDirective($log, iconService) {
```
tbouron: Seems to not be necessary. The `$log` service is not used and the `iconService` is injected on the controller level

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
```diff
@@ -191,3 +188,4 @@ export function entityNodeDirective() {
 
     }
 }
+entityNodeDirective.$inject = ['$log', 'iconService'];
```
tbouron: Same as above

ui-modules/app-inspector/app/components/entity-tree/entity-node.html
```diff
@@ -22,9 +22,12 @@
             <brooklyn-status-icon value="{{entity.serviceState}}" ng-if="entity.serviceState || entity.applicationId"></brooklyn-status-icon>
             <i class="fa fa-2x fa-external-link" ng-if="!entity.serviceState && !entity.applicationId"></i>
             <span>{{entity.name}}</span>
+            <span class="node-icon"><img ng-src="{{ iconUrl }}"/></span>
```
tbouron: Should use the `entity.iconUrl` instead of adding this into the scope, so the icon will be saved on the model which is propagated everywhere.

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
```diff
@@ -130,9 +126,10 @@ export function entityNodeDirective() {
         });
     }
 
-    function controller ($scope, $state, $stateParams) {
+    function controller ($scope, $state, $stateParams, iconService) {
         $scope.isOpen = true;
-
+        iconService.get($scope.entity, true).then(value => $scope.iconUrl = value);
```
tbouron: Should set the `entity.iconUrl` instead of adding this into the scope, so the icon will be saved on the model which is propagated everywhere.

ui-modules/app-inspector/app/views/main/inspect/summary/summary.controller.js
```diff
@@ -51,14 +51,14 @@ export function summaryController($scope, $state, $stateParams, $q, $http, brSna
...
-        vm.entity = response.data;
-        vm.name = response.data.name;
-        vm.error.entity = undefined;
-        observers.push(response.subscribe((response)=> {
+        let set = (response) => {
             vm.entity = response.data;
             vm.name = response.data.name;
             vm.error.entity = undefined;
-        }));
+            iconService.get(response.data, true).then(value => vm.iconUrl = value);
```
tbouron: Should set the `entity.iconUrl` instead of adding this into the scope, so the icon will be saved on the model which is propagated everywhere.

ui-modules/utils/icon-generator/icon-generator.js
```diff
@@ -104,21 +117,49 @@ function IconGenerator(cache, background, margin, size) {
...
+            return new IconService($q, $http, iconGenerator, $log, new SessionsStorageWrapper(CACHE_NAME));
         }]
     }
 }
 
-function IconService($q, $http, iconGenerator, cache) {
+function IconService($q, $http, iconGenerator, $log, cache) {
     this.get = getIcon;
-    function getIcon(id) {
+    function getIcon(entityOrTypeId, doNotAutogenerate) {
```
tbouron: As far as I can see, the there is no called using the `doNotAutogenerate`. It seems to be used only on the pipe.

ui-modules/app-inspector/app/components/entity-tree/entity-node.less
ahgittin: as it's non-growable and the other field is growable, we don't need it at all, quite right

ui-modules/app-inspector/app/components/entity-tree/entity-node.html
ahgittin: the `ng-if` hides the toggle on leaf nodes.  but we want the icons to align so we need a placeholder/wrapper.

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
ahgittin: correct.  also removed the `inject` line

ui-modules/app-inspector/app/components/entity-tree/entity-node.html
ahgittin: nice idea but i think `$scope.entity` can be updated by the caller, so we'd need to watch for it to prevent it being cleared.  it isn't usually the icon which updates, it's things like status, so watching for that and reloading icon seems overkill.

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
ahgittin: yup

ui-modules/utils/icon-generator/icon-generator.js
ahgittin: it's used in `entity-tree.directive.js`:

```
iconService.get($scope.entity, true).then(value => $scope.iconUrl = value);
```

ahgittin:
all addressed

screen-shot:

image

ahgittin:
merging now, as discussed with 1:1 review @tbouron

PR #142, opened 2019-11-18
add an "Open in Composer" button to quick launch

ahgittin:
the effect is to open the template's definition in composer

viz, clicking one of the quick-launch templates show

image

`YAML` opens the in-modal editor (as before), but the new `Open in Composer` takes you to:

image

the location and config are as selected in quick launch, with the full original template


ahgittin:
in quick launch, if any app is tagged `catalog_quick_launch`, then only apps with that tags will be shown.  as before in all cases this only shows those apps marked as templates.

ahgittin:
merging, as discussed with 1:1 review @tbouron

PR #149, opened 2019-12-03
Add jenkinsfile for multibranch job on Jenkins

tbouron:
With this, the multi branch job on Jenkins (to be created) finally works! 🎉


tbouron:
The Jenkins job doesn't use Docker and because some of the binaries required to build the project are missing, the build therefore fails.

Which is pretty ironic as this PR is to fix the situation :)

PS: this builds fines locally

tbouron:
Thanks @kemitix, merging...

PR #150, opened 2019-12-04
add time information to application data to allow ordering launch time

frogfather:
https://issues.apache.org/jira/browse/BROOKLYN-620
Data retrieved from the application api does not contain any time information. 
To allow ordering by time we have to make a further call to entityApi.entityActivities. This returns an array of all activities. We take the startTimeUtc value of the first entry as the application start.

However...

The application data is updated frequently and this erases the time information. To avoid having to make multiple api calls (which causes an annoying flicker on the screen) we store the time information locally in an array of objects. For existing applications we can simply merge this time information with the refreshed data. For new applications, we temporarily set the startTimeUtc to the current time to ensure it's at the top of the list and then call a method to update our local time data.

Screenshot 2019-12-12 at 18 31 16

Also saves sort order preference which resolves 
https://issues.apache.org/jira/browse/BROOKLYN-622

### DIFF
--- a/ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
+++ b/ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
@@ -57,14 +57,8 @@ export function entityTreeDirective() {
 
         let observers = [];
 
-        let timeData = [];
-
         applicationApi.applicationsTree().then((response)=> {
             vm.applications = response.data;
-            vm.applications.forEach(app => {
-                getTimeData(app);
-            });
-
 
             observers.push(response.subscribe((response)=> {
                 response.data
@@ -84,17 +78,7 @@ export function entityTreeDirective() {
                         });
                     });
 
-                //Unless an app has just been loaded the start time information will be held locally in 'timeData'
-                vm.applications = response.data.map(app => {
-                    const appStartData = timeData.find( item => item.applicationId === app.applicationId);
-                    if (!appStartData) {
-                        getTimeData (app)
-                    }
-                    return {
-                        ...app,
-                        startTimeUtc: appStartData ? appStartData.startTimeUtc : Date.now()
-                    }
-                });
+                vm.applications = response.data;
 
                 function spawnNotification(app, opts) {
                     iconService.get(app).then((icon)=> {
@@ -106,24 +90,6 @@ export function entityTreeDirective() {
                     });
                 }
             }));
-            
-        //retrieves start time of the app from the entity api.
-        function getTimeData(app) {
-            //remove entries from timeData relating to apps that have been undeployed
-            timeData = timeData.filter(item => vm.applications.find(app => app.applicationId === item.applicationId))
-            entityApi.entityActivities(app.applicationId, app.applicationId).then( response => {
-                if ((response.data.length === 0) || (timeData.find(item => item.applicationId === app.applicationId))) {
-                    return;
-                };
-                timeData.push({
-                    applicationId: app.applicationId,
-                    startTimeUtc: response.data[0].startTimeUtc
-                });
-                app.startTimeUtc = response.data[0].startTimeUtc
-            }).catch((error) => {
-                console.log(error);
-            });
-            }
         });
 
         $scope.$on('$destroy', ()=> {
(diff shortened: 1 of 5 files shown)


frogfather:
Please re-test this

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
```diff
@@ -43,19 +43,25 @@ export function entityTreeDirective() {
...
         let vm = this;
 
         let observers = [];
 
+        let timeData = [];
+
         applicationApi.applicationsTree().then((response)=> {
             vm.applications = response.data;
+            vm.applications.forEach(app => {
+                getTimeData(app);
```
tbouron: I'm wondering if getting the time of the first activity is really necessary. `applicationApi.applicationsTree()` API returns the list of applications already ordered by age (older first). So theoretically, the only thing you need to is add an index to each application object to be able to sort ascending or descending.

ui-modules/app-inspector/app/components/entity-tree/entity-tree.html
```diff
@@ -16,7 +16,7 @@
   specific language governing permissions and limitations
   under the License.
 -->
-<entity-node ng-repeat="application in vm.applications track by application.id" entity="application" application-id="application.id"></entity-node>
+<entity-node ng-repeat="application in vm.applications | orderBy: '-startTimeUtc' track by application.id" entity="application" application-id="application.id"></entity-node>
```
tbouron: This should not be hardcoded here (always give the choice to the user)
Instead, I think there should be a button to sort by age asc/desc which would trigger the sort. The current sort should probably be remembered as well, either by storing it in a cookie or local storage

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
frogfather: The list returned from applicationsTree isn't in order - In process of this story I launched three empty software processes labelled 1,2,3 and the order returned by the api was not the same as the launch age.

tbouron: @frogfather Hum, the underlying REST API `/v1/applications/fetch` returns consistently the same order on my local environment. So maybe the javascript code does something else somewhere.

ui-modules/app-inspector/app/components/entity-tree/entity-tree.html
frogfather: Good idea - noted

ui-modules/app-inspector/app/components/entity-tree/entity-tree.directive.js
frogfather: @tbouron Odd - before modifying any code the order returned was definitely not the launch order. 
The story was raised because it was noticed that the display order didn't correspond to the launch order.

frogfather:
Please re-test this

ui-modules/app-inspector/app/views/main/main.template.html
```diff
@@ -24,6 +24,11 @@
                     <div class="entity-tree-header">
                         <h2 class="entity-tree-title">Applications</h2>
 
+                        <button class="entity-sort-action" >
+                            <div class="entity-sort-action-toggle" ng-click="ctrl.toggleSortOrder($event)" >
```
tbouron: `ng-click` should be on the button

frogfather: @tbouron changed.

ahgittin:
would it be easier to add the start-time server-side e.g. with a `result.setExtraField("startTimeUtc", ...)` in `ApplicationResource.fromEntity` ?

frogfather:
> would it be easier to add the start-time server-side e.g. with a `result.setExtraField("startTimeUtc", ...)` in `ApplicationResource.fromEntity` ?

It would be a better overall solution - I was quite surprised the application data had no time information assocated with it - but given that we wanted it done this year doing it client side was quicker. Could certainly be redone if one of you is available to pair.

frogfather:
This has now been refactored to use time information returned from the rest server. PR #1076 on brooklyn-server, which adds the time data to the returned data needs to be merged to make this work.

tbouron:
FYI: https://github.com/apache/brooklyn-server/pull/1076 needs to be merged first

PR #154, opened 2019-12-09
add custom onChange directive to correct file uploader bug

frogfather:
Corrects a bug where uploading a file to the catalog via the file picker did not work.
https://issues.apache.org/jira/browse/BROOKLYN-624

Resolves an issue with the getCatalogItemUrl method, which checks for the existence of a string 'org.apache.brooklyn.api.location.Location' in item.supertypes to determine the format of the url:

- Adds a check for the existence of the property supertypes before calling 'includes' on it.
- Checks for the string in item.tags if the supertypes property is not present.

### DIFF
--- a/ui-modules/utils/catalog-uploader/catalog-uploader.js
+++ b/ui-modules/utils/catalog-uploader/catalog-uploader.js
@@ -33,6 +33,7 @@ const MODULE_NAME = 'brooklyn.components.catalog-uploader';
  */
 angular.module(MODULE_NAME, [catalogApi])
     .service('brooklynCatalogUploader', ['$q', 'catalogApi', catalogUploaderService])
+    .directive('customOnChange', customOnChangeDirective)
     .directive('brooklynCatalogUploader', ['$compile', 'brooklynCatalogUploader', catalogUploaderDirective]);
 
 export default MODULE_NAME;
@@ -98,8 +99,8 @@ export function catalogUploaderDirective($compile, brooklynCatalogUploader) {
             element.removeClass('br-drag-active');
         };
 
-        scope.filesChanged = (target)=> {
-            scope.upload(target.files);
+        scope.filesChanged = (event)=> {
+            scope.upload(event.target.files);
         };
 
         scope.upload = (files)=> {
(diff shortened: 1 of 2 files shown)


ui-modules/utils/catalog-uploader/catalog-uploader.js
```diff
@@ -120,7 +121,9 @@ export function catalogUploaderDirective($compile, brooklynCatalogUploader) {
         };
 
         scope.getCatalogItemUrl = (item)=> {
-            return item.supertypes.includes('org.apache.brooklyn.api.location.Location')
+            let itemTraits = item.tags? item.tags.find(item => item.hasOwnProperty("traits")) : {"traits":[]};
+            return (item.supertypes ? item.supertypes : itemTraits.traits)
+                .includes('org.apache.brooklyn.api.location.Location')
```
tbouron: Why is this change for? It doesn't look related to the file uploader

ui-modules/utils/catalog-uploader/catalog-uploader.js
```diff
@@ -208,3 +211,20 @@ export function catalogUploaderService($q, catalogApi) {
...
+        link: function (scope, element, attrs) {
+            element.on('change', x => {
+                var onChangeHandler = scope.$eval(attrs.customOnChange);
+                onChangeHandler(x);
+            });
+            element.on('$destroy', function() {
+                element.off();
+            });
+
+
```
tbouron: Couple of empty lines to remove

ui-modules/utils/catalog-uploader/catalog-uploader.js
frogfather: @tbouron  Strictly not, but it was a minor bug we corrected at the same time. Can make it into a separate PR if required.

tbouron: @frogfather Leaving it there is fine but it would be great if you can explain why this is for

ui-modules/utils/catalog-uploader/catalog-uploader.js
frogfather: Done

ui-modules/utils/catalog-uploader/catalog-uploader.js
frogfather: @tbouron Added an explanation in the description of the PR. Do you also want comments in the code?

PR #156, opened 2019-12-13
minor cleanups with version + NPE

ahgittin:
- if using quick launcher or catalog "deploy", it ignored the version; now it respects the version
- in composer, it could break if a config value was null, now it guards against that
- in composer, name wrapping was only permitted for words, which meant `type.name.xxxxxx` wouldn't wrap; now it can wrap anywhere


frogfather:
@ahgittin LGTM - still waiting for tests to pass. Will merge once done.

PR #157, opened 2019-12-16
Update Jenkinsfile

tbouron:
This does the following:
- add the pipeline trigger on comment only if we are building a PR
- not deploy artifacts when building master. This will be taken care of during the full brooklyn build on master (see Jenkinsfile configuration at the brooklyn root) and should decrease the build time


ahgittin:
LGTM

PR #160, opened 2019-12-19
add relativePath to parent for release make

geomacy:
This adds back the `relativePath` to the parent in the pom, to allow the `make_release_artifacts.sh` to work. This is required to get https://github.com/apache/brooklyn-dist/pull/150 to work.

Without this commit the script breaks like below (abbreviated):

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for org.apache.brooklyn.ui:brooklyn-ui-parent:${revision}: Could not find artifact org.apache.brooklyn:brooklyn-parent:pom:1.0.0 in ...) and 'parent.relativePath' points at wrong local POM @ line 24, column 13
````

@ahgittin this undoes the change https://github.com/apache/brooklyn-ui/commit/f6b88085815bb044c782db1344a5536c8730d01f and probably therefore undoes the fix for embedded builds. What do you think of this? Could we merge this and figure out some other fix for those?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,8 +25,7 @@
         <groupId>org.apache.brooklyn</groupId>
         <artifactId>brooklyn-parent</artifactId>
         <version>1.0.0-SNAPSHOT</version>  <!-- BROOKLYN_VERSION -->
-        <!-- do not link to sibling directory as that breaks builds where this code is embedded, e.g. for branding;
-             we require that brooklyn-server is built first (this is needed anyway for the modularity java code) -->
+        <relativePath>../brooklyn-server/parent/</relativePath>
     </parent>
 
     <groupId>org.apache.brooklyn.ui</groupId>


tbouron:
@geomacy @ahgittin what about adding back the relative path (because Brooklyn needs it) but adding a profile (something like `branded`) which would set it to empty?

tbouron:
Btw, LGTM.

Also, the build error happened to me yesterday, just need to clean the m2 cache but could find the option on Jenkins last night.

geomacy:
> @geomacy @ahgittin what about adding back the relative path (because Brooklyn needs it) but adding a profile (something like `branded`) which would set it to empty?

Can you do that with a profile?

tbouron:
@geomacy I think so. Let me try quickly

tbouron:
@geomacy Seems to works by using a property, referenced in `<relativePath>${relative.path</relativePath>. By default, the property is set to ./brooklyn-server/parent/pom.xml` but is overridden in the `branded` profile to be empty.

tbouron:
retest this please

ahgittin:
nice answer re properties @tbouron -- maven says

> Properties are not supported in the following XPath elements:
> `/project(/parent)?/(groupId/artifactId/version)`

pleasantly surprised that `relativePath` allows this!

i confess i don't understand why we need to set the `relativePath` -- it shouldn't be required if the parent is in the local `.m2` repo, nor if the parent is built prior to the project on the same `mvn` run.  both of those should be the case here, right?

if one of you can explain that that would be helpful, meanwhile i'll check whether the `BrooklynUiParentRelativePath` solution is sufficient for downstream projects, and open a PR building on this if so.

tbouron:
@ahgittin In normal circonstances, it is not needed because as you said, maven is able to fetch the parent pom by the version.

However, when you do change the version when say, you do a release, then the new version (e.g `1.0.0` instead of `1.0.0-SNAPSHOT`) doesn't exist. Maven tries to download it from repositories but canot find it and therefore fails. Whereas if you do have the `<relativePath>` set up, maven falls back to use this during the build, which passes.

ahgittin:
thanks @tbouron -- not 100% clear though, even when cutting a new version, we build `brooklyn-server` before `brooklyn-ui`, so `brooklyn-server/parent` should run and its artifact be available even without relative path.  i trust you that this isn't happening, but i'm wondering if that's a sign of something else not working as it should...

tbouron:
Maybe we do something wrong. What I know is that, with the current state, if you try to follow the release instructions (by changing the version from `1.0.0-SNAPSHOT` to `1.0.0`) the `make_release_artifacts.sh` script fails at the start complaining about this.

tbouron:
Maybe this is an issue with our script @ahgittin but in this case, I don't know how to fix this.

ahgittin:
thanks.  i'll see if there's anything that jumps out in the make without `relativePath` but where i've been testing i think your property idea will work and is probably an improvement.  look out for a merge and new PR in a few minutes.

also FYI from the Javadoc for getRelativePath():

> Get the relative path of the parent pom.xml  file within the check
out. The default value is ../pom.xml. Maven looks for the parent pom
first in the reactor of currently building projects, then in this
location on the filesystem, then the local repository, and lastly in
the remote repo. relativePath allows you to select a different
location, for example when your structure is flat, or deeper without
an intermediate parent pom. However, the group ID, artifact ID and
version are still required, and must match the file in the location
given or it will revert to the repository for the POM. This feature is
only for enhancing the development in a local checkout of that
project.

ahgittin:
> Maven looks for the parent pom first in the reactor of currently building projects

Seems ^ isn't happening so it does need `relativePath` when building a new version.  I suspect our use of `${revision}` is triggering a bug when ^ is resolved.  So +1 to this, and fixing things for downstream subsequently.

PR #161, opened 2019-12-20
require endpoint when provider is azure-arm

frogfather:
Causes the endpoint field to be required when the provider is azure-arm.


ahgittin:
LGTM

PR #162, opened 2019-12-20
Relative path configurable

ahgittin:
extends #160 offering a way for downstream projects to customize relative path if it causes issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
         <groupId>org.apache.brooklyn</groupId>
         <artifactId>brooklyn-parent</artifactId>
         <version>1.0.0-SNAPSHOT</version>  <!-- BROOKLYN_VERSION -->
-        <relativePath>../brooklyn-server/parent/</relativePath>
+        <relativePath>${brooklyn.ui.relativePath.to.brooklyn.server.parent}</relativePath>
     </parent>
 
     <groupId>org.apache.brooklyn.ui</groupId>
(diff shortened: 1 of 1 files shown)


ahgittin:
thanks @tbouron for the suggestion in #160 , implemented here

pom.xml
```diff
@@ -85,6 +84,8 @@
         <brooklyn.version>1.0.0-SNAPSHOT</brooklyn.version><!-- BROOKLYN_VERSION -->
         <build.version>${revision}</build.version>
         <build.name>Apache Brooklyn</build.name>
+        <brooklyn.ui.relativePath.to.brooklyn.server.parent>../brooklyn-server/parent/</brooklyn.ui.relativePath.to.brooklyn.server.parent>
```
tbouron: The name is a bit too much for my taste @ahgittin, I would use a simpler `parent.relative.path`.
But happy to go ahead and merge if it works

ahgittin: assuming this is set higher up in a complex project hierarchy it needs to be namespaced appropriately.  someone else might have a different `parent.relative.path`.  agree it's a little bit silly long but at least it's totally unambiguous. :)

ahgittin:
Seems maven doesn't actually respect a variable in relative path.  I must have been taking some cached or downloaded item when i tested, as when we change the version it breaks.

Also irritating, seems maven won't look in the local reactor model so we need either relative path to work or the upstream item to have been built.

Revert this.

PR #163, opened 2020-01-03
Fixing an issue when building brooklyn-ui for downstream projects

tbouron:
The issue was fixed in version 1.9.0 of the plugin, see: https://github.com/eirslett/frontend-maven-plugin/issues/820#issuecomment-570634363


ahgittin:
LGTM

PR #164, opened 2020-01-08
auto-set a name for the app in quick-launch, and

ahgittin:
apply catalog_quick_launch tag to filtering locations,
if there is a single location select it automatically,
and if these tags are used, suppress the "create a new location" message --
thus making quick-launch suited for a neatly curated simple mode
where a single preferred location is tagged and you can simply click "deploy"
in quick launch without needing to set anything (assuming default config is valid in the blueprint)

### DIFF
--- a/ui-modules/home/app/views/main/deploy/deploy.controller.js
+++ b/ui-modules/home/app/views/main/deploy/deploy.controller.js
@@ -22,6 +22,7 @@ import uiRouter from 'angular-ui-router';
 import brooklynApi from 'brooklyn-ui-utils/brooklyn.api/brooklyn.api';
 import {HIDE_INTERSTITIAL_SPINNER_EVENT} from 'brooklyn-ui-utils/interstitial-spinner/interstitial-spinner';
 import modalTemplate from './modal.template.html';
+import {filterCatalogQuickLaunch} from '../main.controller.js';
 
 const MODULE_NAME = 'states.main.deploy';
 
@@ -54,9 +55,7 @@ export function deployStateController($scope, $state, $stateParams, $uibModal, b
             entitySpec: ['catalogApi', (catalogApi) => {
                 return catalogApi.getBundleType($stateParams.bundleSymbolicName, $stateParams.bundleVersion, $stateParams.typeSymbolicName, $stateParams.typeVersion);
             }],
-            locations: ['locationApi', (locationApi) => {
-                return locationApi.getLocations();
-            }]
+            locations: ['locationApi', locationApi => locationApi.getLocations()],
         }
     });
 
@@ -80,9 +79,17 @@ export function deployStateController($scope, $state, $stateParams, $uibModal, b
 
     function modalController($scope, $location, entitySpec, locations) {
         $scope.app = entitySpec;
-        $scope.locations = locations;
-        // can optionally add: { noEditButton: true, noComposerButton: true }, or pass in URL
-        $scope.args = angular.extend({}, $location.search());
+        $scope.locations = filterCatalogQuickLaunch(locations, (t) => {
+                $scope.usingCatalogQuickLaunchTags = t.length > 0;
+            });
+        
+        // also supports { noEditButton: true, noComposerButton: true }
+        // see quick-launch.js for more info
+        $scope.args = angular.extend({
+                noCreateLocationLink: $scope.usingCatalogQuickLaunchTags
+            },
+            $location.search());
+        
     }
 }
 
--- a/ui-modules/home/app/views/main/deploy/deploy.controller.js
+++ b/ui-modules/home/app/views/main/deploy/deploy.controller.js
@@ -22,7 +22,7 @@ import uiRouter from 'angular-ui-router';
 import brooklynApi from 'brooklyn-ui-utils/brooklyn.api/brooklyn.api';
 import {HIDE_INTERSTITIAL_SPINNER_EVENT} from 'brooklyn-ui-utils/interstitial-spinner/interstitial-spinner';
 import modalTemplate from './modal.template.html';
-import {filterCatalogQuickLaunch} from '../main.controller.js';
+import {filterCatalogQuickLaunch} from '../main.controller.js';  // this really should be handled by angular DI 
 
 const MODULE_NAME = 'states.main.deploy';
 
@@ -80,13 +80,16 @@ export function deployStateController($scope, $state, $stateParams, $uibModal, b
     function modalController($scope, $location, entitySpec, locations) {
         $scope.app = entitySpec;
         $scope.locations = filterCatalogQuickLaunch(locations, (t) => {
-                $scope.usingCatalogQuickLaunchTags = t.length > 0;
+                $scope.usingLocationCatalogQuickLaunchTags = t.length > 0;
             });
         
         // also supports { noEditButton: true, noComposerButton: true }
         // see quick-launch.js for more info
         $scope.args = angular.extend({
-                noCreateLocationLink: $scope.usingCatalogQuickLaunchTags
+                // disable "create location" is admin has configured locations with tags.
+                // called out in docs in https://github.com/apache/brooklyn-docs/pull/299.
+                // a better approach would be to add config on the server to configure this.
+                noCreateLocationLink: $scope.usingLocationCatalogQuickLaunchTags
             },
             $location.search());
         
--- a/ui-modules/home/app/views/main/deploy/modal.template.html
+++ b/ui-modules/home/app/views/main/deploy/modal.template.html
@@ -17,4 +17,4 @@
   under the License.
 -->
 <br-svg type="close" class="modal-close pull-right" ng-click="$dismiss('Dismiss deployment modal')"></br-svg>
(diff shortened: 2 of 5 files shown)


nakomis:
This looks fine, but my knowledge of the UI is pretty limited, so it could probably do with a once over my a more experienced pair of eyes

ui-modules/home/app/views/main/deploy/modal.template.html
```diff
@@ -17,4 +17,4 @@
   under the License.
 -->
 <br-svg type="close" class="modal-close pull-right" ng-click="$dismiss('Dismiss deployment modal')"></br-svg>
-<brooklyn-quick-launch app="app" args="args" locations="locations" callback="$close"></brooklyn-quick-launch>
+<brooklyn-quick-launch app="app" args="args" locations="locations" simpleMode="$scope.usingCatalogQuickLaunchTags" callback="$close"></brooklyn-quick-launch>
```
tbouron: Should call directly `usingCatalogQuickLaunchTags` as the current context is `$scope`

ui-modules/utils/quick-launch/quick-launch.html
```diff
@@ -39,7 +39,8 @@ <h4 class="media-heading">{{app.name || app.symbolicName}}</h4>
             <h3 class="quick-launch-section-title">Name</h3>
             <div class="quick-launch-section-content">
                 <div class="form-group">
-                    <input class="form-control" type="text" ng-model="model.name" ng-disabled="deploying" name="name" placeholder="Choose a name for this application (Optional)" autofocus />
+                    <input class="form-control" type="text" ng-model="model.name" ng-disabled="deploying" name="name" 
+                        placeholder="{{ 'Choose a name for this application (Optional)' }}" />
```
tbouron: It's fine to remove the `autofocus` from this field but at least one in the form should have the `autofocus` directive assigned

ui-modules/home/app/views/main/main.controller.js
```diff
@@ -41,20 +41,29 @@ export const mainState = {
...
-                return applications.filter(application => application.template);
-            });
+            return catalogApi.getTypes({params: {supertype: 'org.apache.brooklyn.api.entity.Application'}}).then(
+                applications => filterCatalogQuickLaunch(applications.filter(application => application.template))
+            );
         }]
     }
 };
 
+export function filterCatalogQuickLaunch(list, callbackForFiltered) {
```
tbouron: Should be added as a filter so that any part of the app can get it through dependency injection instead of calling it directly (which introduce some duplication in the generated code)

ui-modules/home/app/views/main/main.controller.js
```diff
@@ -41,20 +41,29 @@ export const mainState = {
...
+    // the callback is used for clients who wish to adjust their behaviour if tags are used,
+    // eg in deploy.controller where noCreateLocationLink is set on the quick launch if there are tagged locations
+    if (!list) { 
+        list = [];
+    }
+    let tagged = list.filter(i => i && i.tags && i.tags.indexOf("catalog_quick_launch")>=0);
+    if (callbackForFiltered) {
+        callbackForFiltered(tagged, list);
+    }
+    return tagged.length ? tagged : list;
```
tbouron: It doesn't match what the comment says: `In all cases only show those marked as templates`

ui-modules/home/app/views/main/deploy/deploy.controller.js
```diff
@@ -80,9 +79,17 @@ export function deployStateController($scope, $state, $stateParams, $uibModal, b
...
-        // can optionally add: { noEditButton: true, noComposerButton: true }, or pass in URL
-        $scope.args = angular.extend({}, $location.search());
+        $scope.locations = filterCatalogQuickLaunch(locations, (t) => {
+                $scope.usingCatalogQuickLaunchTags = t.length > 0;
+            });
+        
+        // also supports { noEditButton: true, noComposerButton: true }
+        // see quick-launch.js for more info
+        $scope.args = angular.extend({
+                noCreateLocationLink: $scope.usingCatalogQuickLaunchTags
```
tbouron: Why disabling the location creation button if there are quick launch items? It doesn't really make sense to me.

ui-modules/home/app/views/main/deploy/modal.template.html
tbouron: Also, `simpleMode` is not used anywhere, not is a option for the directive `brooklyn-quick-launch`. What is it for exactly?

ui-modules/home/app/views/main/main.controller.js
```diff
@@ -41,20 +41,29 @@ export const mainState = {
...
-                // then only apps with such tags will be shown;
-                // in all cases only show those marked as templates
-                var appsWithTag = applications.filter(application => application.tags && application.tags.indexOf("catalog_quick_launch")>=0);
-                if (appsWithTag.length) {
-                    applications = appsWithTag;
-                }
-                return applications.filter(application => application.template);
-            });
+            return catalogApi.getTypes({params: {supertype: 'org.apache.brooklyn.api.entity.Application'}}).then(
+                applications => filterCatalogQuickLaunch(applications.filter(application => application.template))
```
tbouron: Should register `filterCatalogQuickLaunch` and use the `$filters` service + DI to get it.

ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -38,7 +38,7 @@ export function quickLaunchDirective() {
         scope: {
             app: '=',
             locations: '=',
-            args: '=?',
+            args: '=?', // default: { noEditButton: false, noComposerButton: false, noCreateLocationLink: false, location: null }
```
tbouron: Where this default comes from? It not enforce anywhere nor used.

ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -47,11 +47,21 @@ export function quickLaunchDirective() {
     function controller($scope, $http, $location, brSnackbar) {
         $scope.deploying = false;
         $scope.model = {
-            newConfigFormOpen: false
+            newConfigFormOpen: false,
+            
+            // should never be null, so the placeholder in UI for model.name will never be used;
+            // hence autofocus is disabled
```
tbouron: The name field might be filled up, but the input itself is not disabled, and there are other fields in the form so `autofocus` should be set on one of these.

ui-modules/home/app/views/main/main.controller.js
ahgittin: fixed

ui-modules/utils/quick-launch/quick-launch.js
ahgittin: the values show what the default behaviour according to the code is

ui-modules/utils/quick-launch/quick-launch.js
ahgittin: there's a good chance that the location and the config don't need any change so in this instance autofocussing on anything could be misleading.  arguably we should check if anything is required and make the first required thing autofocus but not sure that's worth the work, and if nothing is required it's unclear to we select "deploy" (even though user might want to change non-required things) or something non-required (which would be confusing).  better IMO to have the tab set just before the first field so it's easy to select using mouse or keyboard - but let's chat.

ui-modules/utils/quick-launch/quick-launch.html
ahgittin: for me using it, it seems best not to have any with `autofocus` -- see below

ui-modules/home/app/views/main/deploy/modal.template.html
ahgittin: quite right, it's not used.  it was to suppress "Create location" when using quick-launch tags (as discussed below).  removed.

ui-modules/home/app/views/main/deploy/deploy.controller.js
ahgittin: the thinking is:

```
+Additionally if any locations have that tag, the "Create Location" option is suppressed,
+as it is assumed the Quick Launch is being used in a curated environment where that option is not meant for end-users.
+(This can be changed through config options in the code, for users who wish to produce their own UI.)
```

which i've added in https://github.com/apache/brooklyn-docs/pull/299/commits/4df37efc19047723ca24e559e398185a4ba08455 (just now).  also changed the var name to make it a little clearer -- `usingLocationCatalogQuickLaunchTags`.  (ie to be clear, this only applies if a _location_ is tagged, not if apps are tagged.)

ui-modules/home/app/views/main/main.controller.js
ahgittin: i don't think it's worth it, as it's not something we expect anyone else to want

ui-modules/utils/quick-launch/quick-launch.js
tbouron: Fair enough, I buy that.

ui-modules/home/app/views/main/deploy/deploy.controller.js
tbouron: Ah got it. Would be good to make it explicitly configurable but we can live with it for now.

PR #166, opened 2020-01-20
use the flattened pom so that revision is set correctly

ahgittin:
brooklyn builds fine without this, but downstream projects get a warning when trying to reference this parent, unless we write the flattened pom to the mvn repo


nakomis:
LGTM; tested locally

PR #167, opened 2020-01-20
Revert "allow relative path to parent to be configured"

ahgittin:
This reverts commit e8dec48eae0b07d1d2918c7315b1090e5c856345.

Seems maven doesn't actually respect a variable in relative path. I must have been taking some cached or downloaded item when i tested, as when we change the version it breaks.

Also irritating, seems maven won't look in the local reactor model so we need either relative path to work or the upstream item to have been built.


nakomis:
LGTM; tested locally

PR #175, opened 2020-06-08
persist sort-by information

frogfather:
Persists whether entities are sorted by bundle or type and the selected sort order.


richardcloudsoft (approved):
Tested, confirmed working.

PR #176, opened 2020-06-09
Remove initial filter if no results

frogfather:
When the spec editor first opens the filters are set to 'Suggested' 'Required' and 'In Use'.
This may result in no results being shown which is confusing to the user as it's not immediately obvious that there are filters in use.

In this situation we add the 'All' filter to show all results. Subsequent operation of the filters is unaffected.


rdowner (approved):
Tested, works as described. Thanks @frogfather !

PR #177, opened 2020-12-02
Use `===` to not only check the value but type

tbouron:
On the spec editor, if an entity has a configuration key of type `integer`, with a `defaultValue` of `0`, then the `Configuration` section displayed `(empty)` because `defaultValue == ''` is true, whereas `defaultValue === ''` is false, which is what we want


tbouron:
Please ignore any TravisCI build errors, it was never meant to be triggered, see https://github.com/apache/brooklyn-ui/pull/179

duncangrant (approved):
LGTM thanks @tbouron

PR #178, opened 2020-12-03
Add support for downloading bundle as ZIP files

tbouron:
This adds a new `Download` button on the bundle page that let a user downloading the current bundle as a ZIP file.

Screenshot 2020-12-03 at 10 20 53

This requires https://github.com/apache/brooklyn-server/pull/1131 to work


tbouron:
Please ignore any TravisCI build errors, it was never meant to be triggered, see https://github.com/apache/brooklyn-ui/pull/179

duncangrant (approved):
LGTM - thanks @tbouron

PR #180, opened 2020-12-07
Make the blueprint composer tabs customisable by downstream projects

tbouron:
Each "tab" is added via a new `TabProvider` and can be removed/updated later on by the application config hook. Each tab is an object with the following properties:
- `title`: the tab title that is going to appears on the tab bar
- `stateKey`: the key of the state to switch to when the tab is clicked on
- `icon`: (optional) the class name for the font-awesome icon to use


iuliana (approved):
Looks good to me.

PR #181, opened 2020-12-14
fields detected as likely "sensitive" now masked in composer by default

ahgittin:
with a little "eye" icon which toggles whether the value is masked (omitted) or shown.
uses same logic as the `sensitive-field` util widget, viz containing "secret" or "password" or "private" etc

image

clicking the slashed eye causes unmasking:

image

when collapsed in unmasked mode it appears like any other widget:

image

in unmasked mode, expanding and clicking the (unslashed) eye re-masks it:

image


duncangrant (approved):
LGTM - thanks

PR #182, opened 2020-12-17
Expose the CodeMirror instance that can be accessed via an event

tbouron:
The YAML editor can now listen for an event with name `brooklyn.component.yaml-editor.cm` and will execute the given callback with the CodeMirror instance. That allows downstream project to use directly `CodeMirror` if they wish too.


iuliana (approved):
LGTM

PR #183, opened 2021-01-06
Switch the linters to be asynchronous.

tbouron:
Because CodeMirror can have only on type of linters when an instance is created (it's either sync or async, cannot have a combination of both) it will restrict at some point what linters downstream projects can define for the composer.

Better to switch the current ones to async so everyone can take advantage of promises.


ahgittin:
LGTM

PR #185, opened 2021-01-12
Use the correct AngularJs module name for blueprint importer

tbouron:
Currently, it will throw an error when the module is started, making it unusable.


tbouron:
FYI, the AngularJs module names for each UI module have been renamed in https://github.com/apache/brooklyn-ui/pull/180. But I made a mistake for the blueprint importer.

All other UI modules work as expected.

nakomis:
LGTM, test pass, and fix tested locally; merging

PR #187, opened 2021-02-03
fix isYamlMode bug and placeholder for auto-detecting yaml state

ahgittin:
the previous refactor left a reference to `isYamlMode` which it removed.  this reinstates that.

we will also want better redirection from the `/yaml?blueprint=xxxx` endpoint links from catalog for those cases where the blueprint is in a non-CAMP format, to allow the correct tab to display.  this adds some placeholder logic for that but does not implement it.


iuliana (commented):
LGTM

PR #189, opened 2021-02-09
Encode Stream Type

duncangrant:
Stream type may be of form "Main artifact:
classpath://scripts/echos/configure.ps1" which is resovled by cxf as a
path


ahgittin:
good fix

PR #196, opened 2021-04-12
SMART-141 Add formatting for WinRM log stream

algairim:

### DIFF
--- a/ui-modules/app-inspector/app/components/stream/stream.directive.js
+++ b/ui-modules/app-inspector/app/components/stream/stream.directive.js
@@ -22,11 +22,11 @@ import template from './stream.template.html';
 const MODULE_NAME = 'inspector.stream';
 
 angular.module(MODULE_NAME, [])
-    .directive('stream', streamDirective);
+    .directive('stream', ["$compile", streamDirective]);
 
 export default MODULE_NAME;
 
-export function streamDirective() {
+export function streamDirective($compile) {
     return {
         template: template,
         restrict: 'E',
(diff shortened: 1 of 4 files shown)


ui-modules/app-inspector/app/components/stream/stream.less
```diff
@@ -49,25 +69,50 @@ stream {
...
+        color: desaturate(@trace-log-color, 20%);
+      }
     }
   }
   
   .fa-arrow-circle-o-down {
     margin-top: 1px;
   }
-
-  .log-autoupdate {
```
algairim, line 64: Replaced with 'fa-spin'.

ui-modules/app-inspector/app/components/stream/stream.directive.js
```diff
@@ -43,19 +43,51 @@ export function streamDirective() {
...
+        $scope.getFormattedItemLogLevel = getFormattedItemLogLevel;
+
+        // CLI XML features
+        $scope.cliXml = false;
+        $scope.cliXmlIdentified = false;
+        $scope.toggleCliXml = toggleCliXml;
+        $scope.isCliXmlSupported = isCliXmlSupported;
+        $scope.cliXmlVerificationRequired = isWinRmStream(); // CLI XML verification is required only when stream is WinRM
+
+        let autoScrollableElement = Array.from(document.getElementsByClassName('auto-scrollable'));
```
ahgittin, line 65: I think we need to restrict to things in the current `$element` -- otherwise an `auto-scrollable` anywhere in the DOM will end up changing the `$scope` of this, no?

ui-modules/app-inspector/app/components/stream/stream.template.html
```diff
@@ -19,14 +19,38 @@
...
-
   <div class="log-actions">
-    <span class="log-autoupdate log-action" ng-click="autoUpdate = !autoUpdate" ng-class="{'active': autoUpdate}" 
-            uib-tooltip="{{ autoUpdate ? 'Disable' : 'Enable' }} auto-update" tooltip-placement="top" tooltip-popup-delay="500" tooltip-append-to-body="true">
-        <i class="fa fa-refresh"></i></span>
-    <span class="log-tail log-action" ng-click="tail = !tail" ng-class="{'active': tail}" 
+
+    <!-- Content filtering -->
+    <span class="log-action" ng-click="isDisplayOther = !isDisplayOther" ng-class="{'active': isDisplayOther, 'hidden': !isFilterContent()}"
+          uib-tooltip="{{ isDisplayOther ? 'Hide' : 'Show' }} other" tooltip-placement="top" tooltip-popup-delay="500" tooltip-append-to-body="true">
```
ahgittin, line 29: Any reason a `.hidden` class is preferred over `ng-show="isFilterContent()` ?  (no big deal)

ahgittin, line 29: would be nice to say in tooltip more than just `Show/Hide` but also what the icon means, how many lines (if easy), and in the case of ther streams if there are other streams

* `Show trace messages (480 lines)`
* `Hide other streams (238 lines): custom1, custom2`

ui-modules/app-inspector/app/components/stream/stream.template.html
```diff
@@ -19,14 +19,38 @@
...
   <pre ng-show="stream.length == 0" class="text-center"><i>(no content)</i></pre>
-
   <div class="log-actions">
-    <span class="log-autoupdate log-action" ng-click="autoUpdate = !autoUpdate" ng-class="{'active': autoUpdate}" 
-            uib-tooltip="{{ autoUpdate ? 'Disable' : 'Enable' }} auto-update" tooltip-placement="top" tooltip-popup-delay="500" tooltip-append-to-body="true">
-        <i class="fa fa-refresh"></i></span>
-    <span class="log-tail log-action" ng-click="tail = !tail" ng-class="{'active': tail}" 
+
+    <!-- Content filtering -->
+    <span class="log-action" ng-click="isDisplayOther = !isDisplayOther" ng-class="{'active': isDisplayOther, 'hidden': !isFilterContent()}"
```
ahgittin, line 28: might be nicer to set in controller `$scope.streams = { debug: { name: "Debug", icon: "fa-bug", visible: true }, ... }` and then here just repeat over it and edit a variable in the object, e.g. `<span ng-repeat="s in Object.values(streams)" ng-click="s.visible = !s.visible" ...>`

(not worth re-doing, just a suggestion)

ahgittin:
looks really good.  one comments which needs considering (use of `document` v `$element`), one more i think worth doing (tooltip)

don't have access to windows so not tested but assuming you've tested i'm happy to merge and correct later if there are glitches

ui-modules/app-inspector/app/components/stream/stream.directive.js
```diff
@@ -67,28 +99,170 @@ export function streamDirective() {
...
+                    formattedItem.isOther = true;
+                }
+
+                // Remove CLI XML string tags for know log levels
+                if (!formattedItem.isOther) {
+                    formattedItem.text = item.replace(/<s s="(.*?)">|\t/gi, '');
+                }
+
+                // Remove CLI XML markers, newlines and replace tabs with spaces
+                formattedItem.text = formattedItem.text.replace(/<\/s>|_x000[a-z0-9]_|\n/gi, '').replace(/\t/g,' ');
```
ahgittin, line 244: will this work for collapsing new lines, or if a new line is in the output?  we split on `\n` to create this loop so feels like if the input is

```
<s s="error">line 1
  line 2
</s>
```

then `line 1` and `line 2` will be other iterations of this loop and so would be classed as `other` ?

ahgittin: i see, we rely on the backend formatting it to single strings for us -- never mind me

ui-modules/app-inspector/app/components/stream/stream.template.html
algairim: I'll keep it as it is, no need to re-do this now.

ui-modules/app-inspector/app/components/stream/stream.directive.js
algairim: Agree, improvement in the new commit.

ui-modules/app-inspector/app/components/stream/stream.template.html
algairim: Changed to ng-show, it is clearer.

algairim: Lines count added into tooltip.

PR #195, opened 2021-04-08
UI efficiency and customisability

ahgittin:
Allow to handle thousands of entities gracefully, and for the composer link to be customized.

PR #197, opened 2021-04-12
Render descriptions using markdown

ahgittin:
Allows descriptions to use markdown, and it formatted as such, with one-line and multi-line support for cases eg where a type has a very long description.

### DIFF
--- a/ui-modules/catalog/app/views/catalog/catalog.template.html
+++ b/ui-modules/catalog/app/views/catalog/catalog.template.html
@@ -69,7 +69,7 @@
         </div>
 
         <div class="col-md-12">
-            <ul class="list-group list-group-bundles" ng-show="state.view === 'bundles'">
+            <ul class="list-group list-group-bundles" ng-if="state.view === 'bundles'">
                 <li ng-repeat="bundle in bundles | filter:state.search | orderBy:state.orderBy.id as filteredBundles track by (bundle.symbolicName + bundle.version)"
                         ng-class="{'expanded': !isCollapsed, 'is-expandable': bundle.types.length > 0}"
                         class="list-group-item bundles-list">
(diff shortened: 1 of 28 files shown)


ahgittin:
Builds on #195

ui-modules/catalog/app/views/catalog/catalog.template.html
```diff
@@ -69,7 +69,7 @@
         </div>
 
         <div class="col-md-12">
-            <ul class="list-group list-group-bundles" ng-show="state.view === 'bundles'">
+            <ul class="list-group list-group-bundles" ng-if="state.view === 'bundles'">
```
algairim, line 72: ng-if removes the whole thing from the DOM tree, I tried it - it is much slower, I would keep ng-show for performance reasons. State can be switched by click, does not make sense to modify DOM tree on every switch.

ui-modules/catalog/app/views/catalog/catalog.template.html
```diff
@@ -165,7 +176,7 @@ <h4>No results matching current filters</h4>
                 </li>
             </ul>
 
-            <ul class="list-group list-group-types" ng-show="state.view === 'types'">
+            <ul class="list-group list-group-types" ng-if="state.view === 'types'">
```
algairim, line 179: Same comment applies here, see above.

algairim (approved):

algairim (approved):

algairim:
I did not know that "Approve" reply approves the whole pull request. I would consider UI performance impact in catalog view switch.

ui-modules/catalog/app/views/catalog/catalog.template.html
ahgittin: If you have a lot of bundles with a lot of types, we _don't_ want them all to render as that is expensive on first load.  I think taking the hit when expanded/collapsed is better.

algairim: I see, makes sense.

PR #202, opened 2021-05-17
use a shared Makefile, and clean/install utils as part of other projects by default

ahgittin:


ahgittin:
not sure why we now need to `npm install` in `utils` ... didn't used to be a problem but now if we don't we see babel complaining about `preset/env`; possibly something in the addition of `marked` in `utils` caused this but it wasn't noticeable except on a clean dev build

algairim (approved):
Looks good.

PR #210, opened 2021-06-01
blueprint composer can show annotations

ahgittin:
based on a `ui-composer-annotation` tag with text value or object value, rendering as markdown

see docs PR to follow immediately

teaser screenshot

image

from code

```
name: Annotation Sample
services:
  - type: aws_s3_bucket
    brooklyn.tags:
      - ui-composer-annotation: A simple default annotation
      - ui-composer-annotation:
          text: >-
            Shown below, yellow text, centered with CSS. Because it's long, scroll bars horizontally and vertically shown when
            needed.
          styleInnerDiv: 'margin: auto; color: yellow;'
          'y': 120
  - type: aws_s3_bucket
    brooklyn.tags:
      - ui-composer-annotation:
          text: |
            ## Big Example
            A **red** tag at _right_, using markdown, in a big box.
          width: 300
          height: 200
          x: 220
          'y': 0
          background: '#ffcccc'
          style: 'font-size: 9px;'
```

### DIFF
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -20,6 +20,8 @@ import * as d3 from 'd3';
 import {PREDICATE_MEMBERSPEC, FIELD} from './model/entity.model';
 import addIcon from '../../img/icon-add.svg';
 import {ISSUE_LEVEL} from './model/issue.model';
+import marked from 'marked';
+import _ from 'lodash';
 
 export function D3Blueprint(container, options) {
     let _svg = d3.select(container).append('svg').attr('class', 'blueprint-canvas');
@@ -157,7 +159,20 @@ export function D3Blueprint(container, options) {
                     y: 75,
                     'xlink:href': addIcon
                 }
-            }
+            },
+            annotation: {
+                foreignObject: {
+                    x: 0,
+                    y: -100,
+                    background:
+                        // "#ffe8b0", // a nice yellow
+                        "#a0c0e0", // a nice light blue
+                    width: 200,
+                    height: 60,
+                    style: "",
+                    styleInnerDiv: "margin: auto 0;",  // center top/down but not left-right by default
+                }
+            },
         },
         transition: 300,
         grid: {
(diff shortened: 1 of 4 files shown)


algairim:
To me it looks like the purpose can be more generic than composer, it could be `ui-annotation` and live in a separate directive.
Other than that it looks useful to have annotations in general, I use notes as an alternative all the time in the graphical editors. It would be good to have a switch to hide them.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -670,6 +685,46 @@ export function D3Blueprint(container, options) {
...
         appendElement(adjunctData.enter(), 'rect', _configHolder.nodes.adjunct.rect);
+
+        let annotationElement = nodeGroup.append('g')
+            .attr('class', 'node-annotation');
+        let annotationData = nodeData.select('g.node-annotation')
+            .selectAll('foreignObject.node-annotation')
+            .data((d)=>{
+                let tags = d.data.metadata && d.data.metadata.get("brooklyn.tags");
+                if (tags) {
+                    let annotations = tags.map(tag => typeof tag === "object" && Object.keys(tag) && Object.keys(tag).length === 1 && tag["ui-composer-annotation"]).filter(x => x);
```
algairim, line 696: Is `filter` at the end required?

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -157,7 +159,20 @@ export function D3Blueprint(container, options) {
...
                     'xlink:href': addIcon
                 }
-            }
+            },
+            annotation: {
+                foreignObject: {
+                    x: 0,
+                    y: -100,
+                    background:
+                        // "#ffe8b0", // a nice yellow
```
algairim, line 168: No toxic colours please!

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
ahgittin: the `map` changes the array to contain the _values_ of the `ui-composer-annotation` key

the `filter(x=>x)` is an idiomatic (if ugly) way to remove `false` and `null` items in set

ahgittin:
there might be scope for other annotations and categories but for now i think err on the side of scoping/namespacing to exactly where they are used

ahgittin:
PS the layers can be used to hide annotations.  i'll add that to the docs.

PR #211, opened 2021-06-03
Login page fixes

duncangrant:
Depends on https://github.com/apache/brooklyn-server/pull/1182

Alert on failed login attempt
Redirect to home if already logged in
Fix issue where after logout and then clicking the return button it may
show the popup instead of redirecting to the login page.

### DIFF
--- a/ui-modules/login/app/views/main/main.controller.js
+++ b/ui-modules/login/app/views/main/main.controller.js
@@ -21,7 +21,7 @@ import uiRouter from 'angular-ui-router';
 import serverApi from 'brooklyn-ui-utils/api/brooklyn/server.js';
 import {HIDE_INTERSTITIAL_SPINNER_EVENT} from 'brooklyn-ui-utils/interstitial-spinner/interstitial-spinner';
 import template from "./main.template.html";
-// import {loginState, loginStateConfig} from "../login/login.controller";
+import modalTemplate from './login-failed.template.html';
 
 const MODULE_NAME = 'states.login';
 
@@ -34,7 +34,7 @@ export const loginState = {
     name: 'login',
     url: '/',
     template: template,
-    controller: ['$scope', '$http', '$window', 'brBrandInfo', loginStateController],
+    controller: ['$scope', '$http', '$window', '$uibModal', 'brBrandInfo', loginStateController],
     controllerAs: 'vm',
     resolve: {}
 };
@@ -43,10 +43,11 @@ export function loginStateConfig($stateProvider) {
     $stateProvider.state(loginState);
 }
 
-export function loginStateController($scope, $http, $window, brBrandInfo) {
+export function loginStateController($scope, $http, $window, $uibModal, brBrandInfo) {
     $scope.$emit(HIDE_INTERSTITIAL_SPINNER_EVENT);
     $scope.getBrandedText = brBrandInfo.getBrandedText;
 
+    let modalInstance = null;
     let loginController = this;
     loginController.error = {};
 
@@ -56,9 +57,13 @@ export function loginStateController($scope, $http, $window, brBrandInfo) {
     }
 
     // If the user is already logged in then redirect to home
-    $http(testAuthReq).then( () => {$window.location.href = '/';} );
+    $http(testAuthReq).then( () => ( $window.location.href = '/' ) );
 
     $scope.login = (user)=> {
+        if (user==null) {
+            openModal();
+            return;
+        }
         var req = {
             method: 'HEAD',
             url: '/v1/server/up/extended',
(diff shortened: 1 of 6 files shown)


ui-modules/login/app/views/main/main.controller.js
```diff
@@ -47,18 +47,28 @@ export function loginStateController($scope, $http, $window, brBrandInfo) {
...
+    let loginController = this;
+    loginController.error = {};
+
+    var testAuthReq = {
+        method: 'HEAD',
+        url: '/v1/server/up/extended'
+    }
+
+    // If the user is already logged in then redirect to home
+    $http(testAuthReq).then( () => {$window.location.href = '/';} );
```
duncangrant, line 59: I think that there is a more angularjs way to redirect but I'm not sure I understand what it is.

ui-modules/login/app/views/main/main.controller.js
```diff
@@ -47,18 +47,28 @@ export function loginStateController($scope, $http, $window, brBrandInfo) {
...
                 'Authorization': 'Basic ' + btoa(user.name +":" + user.password)
             }
         }
 // ui registry metadata
         $http(req)
-            .then(function () {
-                $window.location.href = '/';
-            });
+            .then(() => { $window.location.href = '/'; },
+                () => { $window.alert("Login Failed") ; loginController.error.loginFailed = "Login Failed" });
```
duncangrant, line 72: Alerting works here but feels old / odd and would be better inline but I can't see how to get the loginFailed error to appear on. the form,

ui-modules/login/app/views/main/main.template.html
```diff
@@ -20,6 +20,9 @@
     <div class="row">
         <div class="col-md-4 col-md-offset-4 col-sm-12">
             <form>
+                <p ng-if="error['loginFailed']">
```
duncangrant, line 23: This is a test

ui-modules/utils/server-status/server-status.js
```diff
@@ -65,6 +65,9 @@ export function BrServerStatusDirective() {
                 }else if(response.status === 404) {
                     state = BrServerStatusModalController.STATES.NO_CONNECTION;
                 }else if(response.status === 401 || response.status === 403 ) {
+                    if( response.headers("LOGIN_PAGE")) {
+                        $window.location.href = '/' + response.headers("LOGIN_PAGE");
```
duncangrant, line 69: As before I'm not sure this is the angularjs way

ui-modules/utils/server-status/server-status.js
```diff
@@ -65,6 +65,9 @@ export function BrServerStatusDirective() {
                 }else if(response.status === 404) {
                     state = BrServerStatusModalController.STATES.NO_CONNECTION;
                 }else if(response.status === 401 || response.status === 403 ) {
+                    if( response.headers("LOGIN_PAGE")) {
```
duncangrant, line 68: Is it really bad practice to invent response headers?

algairim: Frankly, I do not know this bit, look at the `$location` service docs https://docs.angularjs.org/guide/$location

ui-modules/utils/server-status/server-status.js
algairim: Frankly, I do not know this bit, look at the `$location` service docs https://docs.angularjs.org/guide/$location

ui-modules/login/app/views/main/main.controller.js
algairim: We do not use alert windows. I would think about modal for it.

ui-modules/utils/server-status/server-status.js
algairim: My vote is to redirect it from the backend.

ui-modules/login/app/views/main/main.controller.js
tbouron: There would be @duncangrant if within the same UI module. But you want to redirect to home, i.e. `/` so this is the right thing to do

ui-modules/login/app/views/main/main.template.html
```diff
@@ -20,6 +20,9 @@
...
+                    fkjldsfdlskjfdslkj
```
tbouron, line 24: Hum, did a cat wrote that? :)

ui-modules/login/app/views/main/main.controller.js
tbouron: Yup, would be nicer to use the `uibModal` service to display a popup rather than the default browser alert.

Also, it is just code style but I find the syntax
```
promise
  .then(() => {})
  .catch(() => {})
```

nicer than the one you are using

ui-modules/utils/server-status/server-status.js
tbouron: I don't think it is. But by convention, there should start with `X_`. 

For instance, I think `X_BROOKLYN_LOGIN_PAGE` would be nicer

tbouron (changes requested):
Thank @duncangrant for the PR. Looks good, just have tiny comments, see below.

Also, I notice that the package-lock.json has changed. This is most likely due to a different version of node/npm you have on your machine when you installed the dependencies. However, you didn't change the dependencies so this should really not be committed here. Would you mind reverting these changes from the PR?

ui-modules/utils/server-status/server-status.js
```diff
@@ -33,13 +33,15 @@ angular.module(MODULE_NAME, [uibModal])
 
 export default MODULE_NAME;
 
+const LOGIN_PAGE_HEADER = "X_BROOKLYN_LOGIN_PAGE";
```
tbouron, line 36: @duncangrant Is that always returned from the API? I'm asking because it is used to build up the link to redirect to

tbouron (commented):
@duncangrant I have one question (see above) but otherwise LGTM!

ui-modules/utils/server-status/server-status.js
tbouron: @duncangrant So if the header is not set, it will redirect to the home page. Is that the expected behaviour?

tbouron (approved):

PR #212, opened 2021-06-04
Made sure the .config.bundle is initialized correctly and thus a corr…

iuliana:
…ect URL is generated when a catalog item  is added from the composer

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -70,24 +70,24 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, $f
         $scope.isNewFromTemplate = () => ($scope.config.itemType !== 'template' && $scope.config.original.itemType === 'template');
         $scope.isUpdate = () => !$scope.isNewFromTemplate() && Object.keys($scope.config.original).length>0;
         $scope.buttonTextFn = () => $scope.config.label || ($scope.isUpdate() && ($scope.config.name || $scope.config.original.name || $scope.config.symbolicName || $scope.config.original.symbolicName)) || 'Add to catalog';
-        $scope.buttonText = $scope.buttonTextFn(); 
-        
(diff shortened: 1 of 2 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -101,9 +101,10 @@ export function saveToCatalogModalDirective($rootScope, $uibModal, $injector, $f
                     $scope.config.version = entity.version || metadata.get('version');
                 }
                 if (!$scope.config.bundle) {
-                    if ($scope.config.symbolicName) {
-                        $scope.config.bundle = $scope.config.symbolicName;
+                    if (!$scope.config.symbolicName) {
+                        $scope.config.symbolicName =  $scope.config.name;
```
algairim, line 105: There is a lot of logic tied up to `symbolicName` and to its absence. It effects the behaviour of c`omposer-blueprint-name-validator` directive a well, the absence of any tests makes it harder to see. I would not initialise `symbolicName` here but assign the `conig.name` to `config.bundle` directly, if `symbolicName` is not initialised.
`symbolicName` is initialized in couple of places already.

iuliana: `symbolicName` is initialized at line 98 already. - only it isn't, that is why the link is broken.

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -142,6 +142,8 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
         saving: false,
         force: false,
     };
+    $scope.catalogBundleId ="";
+    $scope.catalogBundleBase ="";
```
ahgittin, line 146: `""` will generate incorrect URL so not sure it's worth initializing with invalid value, but it should get populated quickly so NBD

ahgittin:
LGTM - merging

PR #216, opened 2021-06-09
Generate the catalog url

iuliana:
... when item is added from the yaml editor. 
@algairim please review.  Thx

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -146,8 +146,6 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
         saving: false,
         force: false,
     };
-    $scope.catalogBundleId ="";
-    $scope.catalogBundleBase ="";
 
     $scope.getTitle = () => {
         switch ($scope.state.view) {
@@ -157,6 +155,15 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
                 return `${$scope.config.name || $scope.config.symbolicName || 'Blueprint'} ${$scope.isUpdate() ? 'updated' : 'saved'}`;
         }
     };
+    $scope.getCatalogURL = () => {
+        switch ($scope.state.view) {
+            case VIEWS.form:
+                return '';
+            case VIEWS.saved:
+                return `/brooklyn-ui-catalog/#!/bundles/catalog-bom-${$scope.config.catalogBundleId}/${$scope.config.version}/types/${$scope.config.catalogBundleBase}/${$scope.config.version}`;
+        }
+    };
+
     $scope.title = $scope.getTitle();
 
     $scope.save = () => {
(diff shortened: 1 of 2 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
```diff
@@ -122,7 +122,7 @@ <h3>You're all set!</h3>
             <p>What would you like to do next?</p>
             <button class="btn btn-default btn-block" ng-click="$dismiss('Continue editing')">Continue to edit this blueprint</button>
             <button class="btn btn-info btn-block" ng-click="$close(REASONS.new)">Create a new blueprint</button>
-            <a class="btn btn-primary btn-block" ng-href="/brooklyn-ui-catalog/#!/bundles/catalog-bom-{{catalogBundleId}}/{{config.version}}/types/{{catalogBundleBase}}/{{config.version}}">View in catalog</a>
+            <a class="btn btn-primary btn-block" ng-href="{{getCatalogURL()}}">View in catalog</a>
```
algairim, line 125: No need to how this button if URL is not available or not valid, consider `ng-if` to hide it until it URL is valid, or at least disable.

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -157,6 +155,15 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
                 return `${$scope.config.name || $scope.config.symbolicName || 'Blueprint'} ${$scope.isUpdate() ? 'updated' : 'saved'}`;
         }
     };
+    $scope.getCatalogURL = () => {
+        switch ($scope.state.view) {
+            case VIEWS.form:
+                return '';
```
algairim, line 161: return null? so it can be used in `ng-if="getCatalogURL()"`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
iuliana: @algairim  I cannot call a function as argument in an `ng-if`. I gave up the `ng-if`  because I noticed the modal title doesn't do it either. Sample: 
` <h3 class="modal-title">{{getTitle()}}</h3>`

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
iuliana: @algairim This url is not shown when the modal is in `form` state so its value is useless here. I can return `null` sure.

algairim: `null` makes sense if you apply any sort of `if` condition to result of `getCatalogURL()`, either implicit or explicit. No need to change it otherwise.

ahgittin:
lgtm

PR #217, opened 2021-06-09
Inherit overflow style for confirmation menu

algairim:


ahgittin:
seems odd to inherit rather than specify what the overflow behaviour here should be.  but i assume this is an improvement!  merging

PR #221, opened 2021-06-11
Logbook UI

jcabrerizo:
UI components to add to the home module a log viewer. Requires https://github.com/apache/brooklyn-server/pull/1188 

image


algairim:
Looks good so far, I can take if from this point.

ahgittin:
Nice, exactly as it should be.  Merge once https://github.com/apache/brooklyn-server/pull/1188 done.

PR #222, opened 2021-06-11
entity lookup in brooklyn service provider with reference ID

zan-mateusz:
Ability to lookup the entity via blueprint service provider using reference ID

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -230,7 +230,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
     }
 
     function lookup(entity, id, any = false) {
-        if (entity._id === id) {
+        if ((entity._id === id) || (entity.id === id)) {
             return entity;
         }
         if (entity.childrenAsMap.has(id)) {


ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -230,7 +230,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
     }
 
     function lookup(entity, id, any = false) {
-        if (entity._id === id) {
+        if ((entity._id === id) || (entity.id === id)) {
```
algairim, line 233: Looks like this lookup is designed for internal `entity._id` only. internal ID is always randomly generated at instantiation of the entity object when `entity.id` is a public value, it is optional and usually is null until assigned. Check if `entity.id` is not `null` first - `... || (entity.id && entity.id === id)`, and then compare with the supplied parameter. I wondering if @ahgittin remembers the history and purpose if this `lookup`. Would be great to add some JSdoc to clarify.

algairim: My point is that `entity._id` is unique, to identify and reference entity model/object in the tree, when `entity.id` is not necessarily unique. Please double check that.

zan-mateusz:
No longer required.

PR #223, opened 2021-06-14
Added fix for catalog 'edit in' links

iuliana:
Edit with YAML editor and Graphical editor now work for multiple yaml plan formats.

### DIFF
--- a/ui-modules/blueprint-composer/app/views/main/main.controller.js
+++ b/ui-modules/blueprint-composer/app/views/main/main.controller.js
@@ -170,6 +170,9 @@ export function MainController($scope, $element, $log, $state, $stateParams, brB
 
     if (yaml) {
         $scope.initialYamlFormat = $stateParams.format;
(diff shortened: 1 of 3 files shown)


algairim:
Looks ok, however, I would check the following condition:
```
if(Array.isArray(edit.type.specList) && edit.type.specList.length > 0 && edit.type.specList[0].format === $stateParams.format) {
    yaml = edit.type.specList[0].contents;
}
```

`Array.isArray()` verifies if it is not null as well.

iuliana:
> Looks ok, however, I would check the following condition:
> 
> ```
> if(Array.isArray(edit.type.specList) && edit.type.specList.length > 0 && edit.type.specList[0].format === $stateParams.format) {
>     yaml = edit.type.specList[0].contents;
> }
> ```
> 
> `Array.isArray()` verifies if it is not null as well.

@algairim you're the JavaScript expert. Updated as requested. Cheers!

algairim (approved):
Looks good now.

PR #227, opened 2021-06-18
Dra & drop improvements

algairim:


algairim:
- [ ] Add new tests for multi-selection mode

ui-modules/blueprint-composer/app/components/designer/designer.directive.js
```diff
@@ -203,38 +204,46 @@ export function designerDirective($log, $state, $q, iconGenerator, catalogApi, b
...
-                    target.addChild(newEntity);
+                    targetEntity.addChild(newEntity);
                 }
                 blueprintService.refreshEntityMetadata(newEntity, EntityFamily.ENTITY).then(() => {
+                    container.dispatchEvent(new CustomEvent('new-entity-created', {
+                        detail: {
+                            nodeId: newEntity._id,
+                            parentId: targetEntity._id
+                        }
+                    }));
```
algairim, line 223: Consistent with 'move-entity' event.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -1097,25 +1097,53 @@ export function D3Blueprint(container, options) {
...
      * @param {String} id The ID of a node to draw confirmation menu for.
      * @param {String} message The confirmation message.
-     * @param {Array.<String>} choices The confirmation choices.
-     * @param {Function} callback The confirmation callback with a boolean parameter, true if confirmed (Yes),
-     *        false if denied (No), null otherwise (ignored).
+     * @param {Array.<String|Object>} choices The confirmation choices as plain strings or objects that implement toString
+     *        to display choice options. Use objects with distinct *.id value for choices that return same toString value.
+     * @param {Function} callback The callback with confirmed choice in the single-selection mode, or with {Array.<String>}
+     *        of confirmed choices in the multi-selection mode.
+     * @param {Boolean} isMultiSelection Indicates if confirmation selection is multi-selection, single-selection is default.
```
ahgittin, line 1142: should `default` of type `<String|Object>` (or `Array` thereof for `isMultiSelection`) be a parameter?

ahgittin: (so callers can specify what is initially selected?)

ahgittin:
looks great, but for one comment on allowing default values to be passed in

ahgittin:
fine to merge as is but will wait for reply or user-click-elsewhere if wanted in this PR

algairim:
Agreed to implement "if user clicks elsewhere - the menu goes away" in a separate PR.

ahgittin:
works great.  no need to allow defaults to be configured just yet -- that can come.  some minor errors and remaining tasks:

* popup goes away if somewhere else is clicked
* issue w z-order meaning parts of the button are not clickable in some cases
* issue w <path attribute "d" incomplete when nodes are moved
* issue on safari where buttons don't appear
* issue after relationship created that the arc doesn't appear

but merging as it's so good; remaining tasks to be addressed/considered in follow-up

ahgittin:
there are known bugs with foreignobject on safari which appear to be causing the problems there:
* https://bugs.webkit.org/show_bug.cgi?id=90738 (has patch)
* https://bugs.webkit.org/show_bug.cgi?id=23113 (depends on above; might be fixed by patch?)

as there are recent patches i think best thing to do is wait for them to land in safari and not worry about those issues for now.  it would need a major rework to abandon foreignobject and is not worth it IMO.

ahgittin:
the z-order / button clickability bug goes away if we reverse the order of the nodes, eg d3-blueprint.js line 549 add `.reverse()`:

```
_d3DataHolder.nodes = root.descendants().reverse();
```

this is because svg layers things in the order they are listed.  so without this the bottom row of buttons is logically on top of its parents `g` element.  `foreignObject` seems to transcend this in some ways but not completely.  painting the things bottom-first means the foreignObject is above the children nodes by all modes of reckoning.

(unfortunately this doesn't fix any of the safari issues, and i've noticed one more nasty one:  sometimes the tab can hang trying to render the popup.  only solution i know of is to kill and restart safari!  even stronger evidence it's a nasty bug in safari, so not worth _our_ trying to fix it, unless we stumble on an easy fix, but hopefully will be fixed in safari/webkit soon!)

PR #230, opened 2021-06-21
Fix/smart 188

jathanasiou:
Using derived properties and with scripted updates based on `state.view` instead of functions inside the View.


iuliana (approved):
LGTM.

iuliana (approved):
LGTM

PR #232, opened 2021-06-23
Draw relationship path labels with textPath, if supplied

algairim:

### DIFF
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -806,7 +806,7 @@ export function D3Blueprint(container, options) {
                 .attr('hidden', (d) => (d.label ? null : ''))
                 .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
                 .attr('startOffset', '59%')
-                .html((d) => (' ' + d.label + ' '));
+                .html((d) => (d.label ? ' ' + d.label + ' ': null));
 
         // Draw the transition
         relationData.transition()
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -786,9 +786,11 @@ export function D3Blueprint(container, options) {
             .attr('from', (d)=>(d.source._id))
             .attr('to', (d)=>(d.target._id));
 
+
         // Draw the relationship label that follows the path, somewhere in the middle.
         // NOTE `textPath` DECREASES THE UI PERFORMANCE, USE LABELS WITH CAUTION.
-        relationDataEntered.insert('text') // Add text layer of '&#9608;'s to erase the area on the path.
+        let relationDataLabelsEntered = relationDataEntered.filter(d => d.label);
+        relationDataLabelsEntered.insert('text') // Add text layer of '&#9608;'s to erase the area on the path.
             .attr('dominant-baseline', 'middle')
             .attr('text-anchor', 'middle')
             .attr('font-family', 'monospace')
@@ -798,7 +800,7 @@ export function D3Blueprint(container, options) {
                 .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
                 .attr('startOffset', '59%') // 59% roughly reflects `middle of the arch` minus `node radius`.
                 .html((d) => (d.label ? '&#9608;'.repeat(d.label.length + 2) : null));
-        relationDataEntered.insert('text') // Add label text on top of '&#9608;'s which is on top of the path.
+        relationDataLabelsEntered.insert('text') // Add label text on top of '&#9608;'s which is on top of the path.
             .attr('dominant-baseline', 'middle')
             .attr('text-anchor', 'middle')
             .attr('font-family', 'monospace')
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -796,7 +796,6 @@ export function D3Blueprint(container, options) {
             .attr('font-family', 'monospace')
             .attr('fill', '#f5f6fa') // colour of the canvas
                 .insert('textPath')
-                .attr('hidden', (d) => (d.label ? null : ''))
                 .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
                 .attr('startOffset', '59%') // 59% roughly reflects `middle of the arch` minus `node radius`.
                 .html((d) => (d.label ? '&#9608;'.repeat(d.label.length + 2) : null));
@@ -805,7 +804,6 @@ export function D3Blueprint(container, options) {
             .attr('text-anchor', 'middle')
             .attr('font-family', 'monospace')
                 .insert('textPath')
-                .attr('hidden', (d) => (d.label ? null : ''))
                 .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
                 .attr('startOffset', '59%') // 59% roughly reflects `middle of the arch` minus `node radius`.
                 .html((d) => (d.label ? ' ' + d.label + ' ': null));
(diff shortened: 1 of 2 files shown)


algairim:
The label text strictly follows the arch, it can be displayed upside-down, depends on arch position. Ideas on how to address that are welcome.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -776,11 +776,39 @@ export function D3Blueprint(container, options) {
...
+        relationDataEntered.insert('text') // Add text layer of '&#9608;'s to erase the area on the path.
+            .attr('dominant-baseline', 'middle')
+            .attr('text-anchor', 'middle')
+            .attr('font-family', 'monospace')
+            .attr('fill', '#f5f6fa')
+                .insert('textPath')
+                .attr('hidden', (d) => (d.label ? null : ''))
+                .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
+                .attr('startOffset', '59%')
+                .html((d) => (d.label ? '&#9608;'.repeat(d.label.length + 2) : null));
```
algairim, line 800: This step might look redundant, however, this is how middle part of the path is erased on the label text:
Screenshot 2021-06-23 at 09 55 22

algairim, line 800: This step might look redundant, however, this is how middle part of the path is erased for the label text:

Screenshot 2021-06-23 at 09 59 42

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -776,11 +776,39 @@ export function D3Blueprint(container, options) {
...
+        relationDataEntered.insert('text') // Add label text on top of '&#9608;'s which is on top of the path.
+            .attr('dominant-baseline', 'middle')
+            .attr('text-anchor', 'middle')
+            .attr('font-family', 'monospace')
+                .insert('textPath')
+                .attr('hidden', (d) => (d.label ? null : ''))
+                .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
+                .attr('startOffset', '59%')
+                .html((d) => (' ' + d.label + ' '));
```
ahgittin, line 809: does this get html-escaped?  try with a label eg `<b>not bold</b>`.  we have `import _ from lodash` available so can use that if we need to escape it.

also do we want a check `(d.label ? ' ' + d.label + ' ' : '')` here?  (i know it's hidden but still feels like good practice.)

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -776,11 +776,39 @@ export function D3Blueprint(container, options) {
...
+                .attr('hidden', (d) => (d.label ? null : ''))
+                .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
+                .attr('startOffset', '59%')
+                .html((d) => (d.label ? '&#9608;'.repeat(d.label.length + 2) : null));
+        relationDataEntered.insert('text') // Add label text on top of '&#9608;'s which is on top of the path.
+            .attr('dominant-baseline', 'middle')
+            .attr('text-anchor', 'middle')
+            .attr('font-family', 'monospace')
+                .insert('textPath')
+                .attr('hidden', (d) => (d.label ? null : ''))
```
ahgittin, line 806: instead of hiding is there a way to filter the `relationDataEntered` based on the value of `d.label` -- so we don't generate this block at all?  (i assume empty text paths don't contribute much to efficiency but it might help some with large blueprints.)

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: No need to use bold, spaces are required to match the length of underneath background, created in the step above `.html((d) => (d.label ? '&#9608;'.repeat(d.label.length + 2) : null));`

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -776,11 +776,39 @@ export function D3Blueprint(container, options) {
...
+                .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
+                .attr('startOffset', '59%')
```
ahgittin, line 808: why `59%` ?  seems curious and arbitrary.  at minimum deserves an explanation.  i noticed some of the long labels seemed to go close to the end of the arc.  if there is a way to actually center it that would be nicer -- but if that's hard (which i assume it is) than a heuristic magic number such as this makes sense.

agree it looks nicest if it's approximately centered in the usual short-text case.  starting the text near the source of the arc would make more logical sense if it has to be left-aligned but definitely wouldn't be as aesthetically pleasing.  (so what you've done might be the best reasonable solution.)

ahgittin:
looks cool.  a few minor comments.  not tested yet -- there might be performance implications but we can address those later (and users can disable these labels if they wish).

ahgittin:
is it worth adding labels for the config-key relationships?

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: It is an heuristic number :), arch starts at the centre of the source node but ends at target's perimeter. 59% roughly reflects `middle of the arch` minus node `radius`. I'll add a comment in the code.

algairim:
> is it worth adding labels for the config-key relationships?

Yes, I would add those for config-key relationships, however, we need to decide what to add as a label. Is it just a name of the config key?

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: I need help with this, I was looking how to achieve this, `hidden` attribute is a workaround.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Null-check added in 66699f208f11c34192d7a7cc71cdd2137dfa88e9

ahgittin:
@algairim yes, the name of the config key is the right thing to show for config i think

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -776,11 +776,39 @@ export function D3Blueprint(container, options) {
...
+        relationDataEntered.insert('path')
             .attr('class', 'relation')
+            .attr('id', (d)=>(d.source._id + '-' + d.target._id))
             .attr('opacity', 0)
             .attr('from', (d)=>(d.source._id))
             .attr('to', (d)=>(d.target._id));
+
+        // Draw the relationship label that follows the path, somewhere in the middle.
+        // NOTE `textPath` DECREASES THE UI PERFORMANCE, USE LABELS WITH CAUTION.
+        relationDataEntered.insert('text') // Add text layer of '&#9608;'s to erase the area on the path.
```
ahgittin, line 791: what about

```
let relationDataLabelsEntered = relationDataEntered.filter(d => d.label)
```

then of course change line above and 801 below to

```
relationDataLabelsEntered.insert('text') ...
```

algairim: This works, fdc24e503da10ea2f65ce4f06e3bda7e26dfe6cb

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Addressed in fdc24e503da10ea2f65ce4f06e3bda7e26dfe6cb

algairim:
> @algairim yes, the name of the config key is the right thing to show for config i think

@ahgittin, we can add labels to config keys in a separate PR, requires a little bit of work.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Removed null-check in 8d4f330fbd6c724caf36cc440fff867147a7c7c8, filter addresses that.

algairim:
Added levels for config keys of type Object b8daa4ec8eaec3e1b733afca194f82d8511a271c, others TODO.

algairim:
> Added levels for config keys of type Object [b8daa4e](https://github.com/apache/brooklyn-ui/commit/b8daa4ec8eaec3e1b733afca194f82d8511a271c), others TODO.

Addressed others.

ahgittin:
Looks great.  For future reference `Object.entries(entity.config).forEach( ([key, config]) => ... )` would be simpler.  But merging!

PR #233, opened 2021-06-23
Make relationship path and label selector configurable

algairim:


ui-modules/blueprint-composer/app/views/main/main.less
```diff
@@ -87,6 +87,9 @@
     text-align: center;
     padding-top: 7px;
   }
+  .dropdown-menu {
+    width: inherit;
```
algairim, line 91: This adjusts the menu width to the widest item in the list.

PR #234, opened 2021-06-25
Enhancement/catalog better overrides support

jathanasiou:
Code cleanup and enhancements to work better with overrides. Added logic enables
- support for async post-processing for YAML plan generation
- Hard flag to hide YAML text editor button
- no impact to default behaviour


iuliana (approved):
I build Brooklyn with this changes and I am satisfied it does not modify the original functionality. Only improves the Javascript syntax.(a lot)

PR #236, opened 2021-06-30
Improved syntax of the updateBundleConfig method and...

iuliana:
also fixed it in the process. `scope` was given as a parameter but `$scope` was used in the method body.


algairim:
Looks good.

iuliana:
I'll postpone merging this until another change depending on this is ready.

PR #237, opened 2021-07-01
Switch relationships with path and label selector rather than group selector

algairim:
Switch relationships with path and label selector rather than whole relationship group, relationships can belong to sub-groups to be managed separately

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -93,8 +93,8 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
                     source: entity,
                     target: relation.entity,
                     label: relation.name,
-                    pathSelector: 'relation', // the CSS class for path
-                    labelSelector: 'relation-label', // the CSS class for label
+                    pathSelector: 'relation-selector', // the CSS class for path
+                    labelSelector: 'relation-selector', // the CSS class for label
                 };
             });
         }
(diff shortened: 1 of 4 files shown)


ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -93,8 +93,8 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
                     source: entity,
                     target: relation.entity,
                     label: relation.name,
-                    pathSelector: 'relation', // the CSS class for path
-                    labelSelector: 'relation-label', // the CSS class for label
+                    pathSelector: 'relation-selector', // the CSS class for path
+                    labelSelector: 'relation-selector', // the CSS class for label
```
iuliana, line 97: `pathSelector` and `labelSelector` both  seem to have the same value. Is this intended?

iuliana: Nevermind, I noticed there is the only one css class for this purpose.

iuliana (commented):
LGTM

iuliana:
Waiting for build to pass then this can be merged.

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
algairim: Yes, selectors are the same for arcs and labels for Brooklyn relationships, however, they are can be different in the downstream versions when selecting label and arc can be managed differently.

PR #242, opened 2021-07-09
Remove a extra redirection when navegating using the menu and the mai…

jcabrerizo:
Without the extra `/` added the browser make a attempt go open the url and after that a second success attempt to the module:
image

By adding slash only one request is needed.


iuliana (approved):
LGTM

PR #245, opened 2021-07-12
Fix/quick launch dialog

jathanasiou:
Fixes to deal with certain cases:

- Pinned config attributes without defaultValue crashing the YAML serializer
- Properly using Location's `id` instead of entire object for predefined/uploaded Location items
- YAML Editor note about changes being lost when pressing "Back"
- Minor code syntax improvements


ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -178,7 +185,8 @@ export function quickLaunchDirective() {
             }
         }
 
-        function addNewConfigKey() {
+        function addNewConfigKey({ name, type }) {
+            // console.log('$item', {name, type })
```
algairim, line 189: To remove.

ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -147,23 +144,33 @@ export function quickLaunchDirective() {
...
+                            if ($scope.callback) {
+                                $scope.callback.apply({}, [{state: 'SUCCESS', data: response.data}]);
+                            } else {
+                                brSnackbar.create('Application Deployed');
+                            }
+                            $scope.deploying = false;
+                        })
+                        .catch(({ data }) => {
+                            // handling API error
+                            handleDeployError({ message: data.message });
```
algairim, line 160: Can be just `handleDeployError(data)`?

jathanasiou: It seems like it could but we're essentially dealing with an http call error and having it mimick a generic `throw 'my message` `Error` object structure. So, it might be best to leave for transparency? I don't mind either way.

ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -276,7 +281,7 @@ export function quickLaunchDirective() {
                         newApp[BROOKLYN_CONFIG] = newConfig;
                     }
                 }
-                
+
                 // prefer to use the actual yaml input, but if it's not possible
                 let tryMergeByConcatenate =
                     Object.keys(newApp).length ?
```
algairim, line 287: Catch block might be required for the Promise in `showEditor` function. Find below.

ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -43,22 +46,20 @@ export function quickLaunchDirective() {
         },
         controller: ['$scope', '$http', '$location', 'brSnackbar', 'brBrandInfo' , 'quickLaunchOverrides', controller]
     };
-
+f
```
algairim, line 49: to remove `f`

algairim:
Refactoring is good. One more Promise possibly to add catch block for.

ui-modules/utils/quick-launch/quick-launch.js
jathanasiou: Added a simple `console.error` since the typical `model.deployError` prop serves a different purpose and this case is controlled enough that it should generally not be able to fail (Form > yaml text generation). Further ideas would be welcome though.

algairim: Agree, good to have log at least.

iuliana (approved):
LGTM

PR #249, opened 2021-07-15
Logbook widget navigation improvements

algairim:


algairim:
Screenshot 2021-07-16 at 07 39 48

algairim:
Log level colors match with the streams in app-inspector, for consistency.

ui-modules/utils/logbook/logbook.js
```diff
@@ -39,168 +39,249 @@ export function logbook() {
...
+                        scrollToMostRecentLogEntry();
+                    } else if (datetimeToScrollTo && datetimeToScrollTo >= $scope.logEntries[0].datetime) {
+                        scrollToLogEntryWithDateTime(datetimeToScrollTo);
+                    }
+                }
+
+                // Re-set marker for search parameters changes after auto-scroll.
+                isNewQueryParameters = false;
+
+                if ($scope.logEntries.length === 0) {
```
jathanasiou, line 281: could be an `else { ..}` to the previous if for brevity

ui-modules/utils/logbook/logbook.js
```diff
@@ -39,168 +39,249 @@ export function logbook() {
...
-        });
-
-        vm.queryTail = () => {
-            let autoUpdate = !$scope.autoUpdate; // Calculate new value.
-
-            if (autoUpdate) {
-                $scope.isAutoScroll = true;
-                resetQueryParameters();
-                doQuery();
+        let autoScrollableElement = Array.from($element.find('pre')).find(item => item.classList.contains('auto-scrollable'));
```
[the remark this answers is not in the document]
algairim: Element object is required here to perform a scroll action on entire log section, not just on a single log entry (log line), when needed. I take any suggestion(s) on how to perform scroll action by using data model, instead of searching element in the DOM.

ui-modules/utils/logbook/logbook.js
```diff
@@ -39,168 +39,249 @@ export function logbook() {
...
+            doQuery();
+        };
+
+        /**
+         * Converts log entry to string.
+         *
+         * @param {Object} entry The log entry to convert.
+         * @returns {String} log entry converted to string.
+         */
+        vm.covertLogEntryToString = (entry) => {
```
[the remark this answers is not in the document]
algairim: Good catch.

ui-modules/utils/logbook/logbook.js
algairim: No, `isNewQueryParameters` must be reset before that, regardless of query result.

ui-modules/utils/logbook/logbook.js
```diff
@@ -39,168 +39,249 @@ export function logbook() {
...
+                outputLine.push(entry.message);
+
+            return outputLine.join(' ');
+        };
+
+        /**
+         * Caches the datetime of the first item in the visible area of the query result.
+         */
+        function cacheDatetimeToScrollTo() {
+            let element = Array.from($element.find('pre')).find(item => item.offsetTop > (autoScrollableElement.scrollTop + autoScrollableElement.offsetTop - 1));
```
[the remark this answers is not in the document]
algairim: Same as above, I'm open to ideas on how to avoid parsing the DOM.

ui-modules/utils/logbook/logbook.js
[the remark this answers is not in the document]
algairim: This is addressed in 3cd18d2f7f1e32424040ea1a88654f1c63c88ebe now

ui-modules/utils/logbook/logbook.js
```diff
@@ -39,168 +39,235 @@ export function logbook() {
...
             }
 
+            cacheDatetimeToScrollTo();
+
             logbookApi.logbookQuery(params, true).then((logEntries) => {
 
-                if ($scope.isLatest && $scope.logEntries.length !== 0) {
-                    if (logEntries.length > 0) {
+                // Assign unique IDs for new log entries.
+                logEntries.forEach(item => item.id = generateLogEntryId());
```
algairim, line 230: Looks like we need proper IDs assigned in the backend data model to make it reliable, confirmed with @jathanasiou

jathanasiou:
# Log items consistency

It seems that the way `logEntries` are being retrieved from the server API right now can be inconsistent.

Using `timestamp` values to define a range the entries can lead to overlaps between consecutive queries. Timestamps can be **relative** and due to the nature of async operations such as this, inappropriate data sets may be sent to the client.


    query-A (from 16:21:0005 to 16:21:0010)
        client request sent at 16:21:0010
        server receives request at 16:21:0012 (or later) --> log entries may have been generated between 0010 and 0012
        client gets response at 16:21:0013 (or later)
    
    query-B (from 16:21:0010 to 16:21:0015)
        // same situation, overlap and transfer delays can cause unexpected behaviour and displayed items
        // the client could receive the same entry twice, or completely miss a log entry

----
## Suggestion: Use absolute reference IDs per entry

The server should be tagging each log entry with a unique UUID (lots of libraries can do that reliably). After the first query, the client can instead use the ID in the last of (ordered) entry results in the most recent query response to request only entries newer than that one.

See also https://github.com/apache/brooklyn-ui/pull/249#pullrequestreview-708378520

# DOM manipulation

Examining elements via `$element` is far from ideal, but somewhat necessary when DOM-related properties (scroll position inside parent) are needed. Once the issue of entry consistency is resolved, it might be easier to evaluate the mechanism for keeping older lines in "focus" via specific scroll positions.

ui-modules/utils/logbook/logbook.js
```diff
@@ -210,36 +285,30 @@ export function logbook() {
...
+         * @returns {number|undefined} The UTC timestamp.
+         */
+        function getUtcTimestamp(date) {
+            const timezoneOffsetMs = new Date().getTimezoneOffset() * 60 * 1000;
+            if (date instanceof Date) {
+                return date.valueOf() - timezoneOffsetMs;
+            } else if (typeof date === 'number') {
+                return date - timezoneOffsetMs;
+            } else {
+                return undefined;
```
jathanasiou, line 310: Return undefined is a needless statement for JS functions.

algairim: I would prefer to keep it explicit.

ahgittin:
LGTM

PR #250, opened 2021-07-16
Make the timeAgoFilter stateful so it keeps updating the display

tbouron:
By default, AngularJS does not reevalute an expression if the left part doesn't change. This is exactly what happens with the `timeAgoFilter`: the timestamp never changes obviously and therefore, the display keeps displaying the evaluated value at the moment of the first page display.

Marking the filter as `stateful` forces AngularJS to reevaluate when the scope changes.

By doing this, the task list (App Inspector) now updates the started time column on the fly

### DIFF
--- a/ui-modules/app-inspector/app/components/task-list/task-list.directive.js
+++ b/ui-modules/app-inspector/app/components/task-list/task-list.directive.js
@@ -84,11 +84,15 @@ export function taskListDirective() {
 }
 
 export function timeAgoFilter() {
-    return function (input) {
+    function timeAgo(input) {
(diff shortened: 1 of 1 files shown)


iuliana (approved):

algairim:
It does not change anything. Is it?

algairim:
Okay, just managed to recreate the old behaviour. The time was not refreshed on its own, now it is working correctly. Ready to merge.

algairim (approved):

tbouron:
Thanks @iuliana @algairim, merging this

PR #252, opened 2021-07-19
Logbook widget in app inspector

algairim:
Logbook widget has been enhanced with multi-phrase search. `searchId` is now a minimum phrase to match with, if supplied in the directive arguments.
Screenshot 2021-07-19 at 18 38 50

### DIFF
--- a/ui-modules/utils/logbook/logbook.js
+++ b/ui-modules/utils/logbook/logbook.js
@@ -40,7 +40,7 @@ export function logbook() {
         let vm = this;
         let refreshFunction = null;
         let autoScrollableElement = Array.from($element.find('pre')).find(item => item.classList.contains('auto-scrollable'));
-        let isNewQueryParameters = true; // Fresh start, new parameters!
+        let queryParametersChanged = 1; // Fresh start, new parameters!
         let datetimeToScrollTo = '';
 
         // Set up cancellation of auto-scrolling down.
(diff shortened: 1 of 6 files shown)


algairim:
Couple of UX questions:
* Do we want to preform an initial single query or start auto-query with default search parameter automatically (It is no at the moment)?
* Would it be better to add activity ID as part of logbook header, so that it will be explicit and imply that logbook is for this activity ID only, meaning some filter applied already. Like the following:
Screenshot 2021-07-20 at 09 24 56

algairim:
Depends on https://github.com/apache/brooklyn-server/pull/1208

algairim:
Added logbook specific in parentheses.
Screenshot 2021-07-20 at 14 32 58

algairim:
Same for activities:
Screenshot 2021-07-20 at 14 38 23

ui-modules/utils/logbook/logbook.js
```diff
@@ -91,8 +91,6 @@ export function logbook() {
         $scope.fieldsToShow = ['timestamp', 'class', 'message']
         $scope.logFields = [
             {name: 'Timestamp',   value: 'timestamp',  selected: true},
-            {name: 'Task ID',     value: 'taskId',     selected: false},
-            {name: 'Entity IDs',  value: 'entityIds',  selected: false},
```
algairim, line 90: @jcabrerizo, there is no task ID or entity ID in logs.

jcabrerizo: Have you tried to invoke an effector? Is how I teste it and the taks Id was there

algairim: Ok, I see now, reverting back.
Screenshot 2021-07-20 at 19 19 59

jcabrerizo (approved):
LGTM @algairim

PR #241, opened 2021-07-08
Add support for an ancestor quick fix

tbouron:
This will check if an entity has implicit inherited config keys. If this is the case, then warnings will be created for these config keys. Additionally, a quick fix will be available to automatically set the implicit inherited config key `x` to `$brooklyn:parent().config("x")`

Here it is in action:
quick-fix-ancestor

PR #255, opened 2021-07-21
Support multiple labels on the same relationship path

algairim:
Screenshot 2021-07-21 at 12 15 21

### DIFF
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -184,8 +184,8 @@ export function D3Blueprint(container, options) {
                 }
             },
             relationship: {
-                pathSelector: 'relation-arc',
-                labelSelector: 'relation-label',
+                pathClass: 'relation-arc',
+                labelClass: 'relation-label',
                 labelBackgroundColor: '#f5f6fa',
                 labelsToDisplay: 2,
                 labelOffsetPx: -12
@@ -808,7 +808,7 @@ export function D3Blueprint(container, options) {
         let relationArcs = _relationArcs.selectAll('.relation').data(arcsData, (d) => getPathId(d));
 
         relationArcs.enter().insert('path')
-            .attr('class', 'relation ' + _configHolder.nodes.relationship.pathSelector)
+            .attr('class', 'relation ' + _configHolder.nodes.relationship.pathClass)
             .attr('id', (d) => getPathId(d))
             .attr('opacity', 0)
             .attr('from', (d) => (d.source._id))
@@ -848,7 +848,7 @@ export function D3Blueprint(container, options) {
         // Draw relationship labels that follow paths, somewhere in the middle of paths.
         // NOTE <textPath/> DECREASES THE UI PERFORMANCE, USE LABELS WITH CAUTION.
 
-        _relationLabels.selectAll('.' + _configHolder.nodes.relationship.labelSelector).remove(); // Re-draw labels every time, required to refresh changes.
+        _relationLabels.selectAll('.' + _configHolder.nodes.relationship.labelClass).remove(); // Re-draw labels every time, required to refresh changes.
 
         // Group unique labels per path.
         let labelsPerPath = {};
@@ -880,7 +880,7 @@ export function D3Blueprint(container, options) {
             return labelIndex > 0 ? labelIndex * _configHolder.nodes.relationship.labelOffsetPx : 0;
         };
 
-        let relationLabelsEntered = _relationLabels.selectAll('.' + _configHolder.nodes.relationship.labelSelector)
+        let relationLabelsEntered = _relationLabels.selectAll('.' + _configHolder.nodes.relationship.labelClass)
             .data(Object.values(labelsData)).enter();
 
         // Returns label text for up to 3 labels at the same path.
@@ -895,7 +895,7 @@ export function D3Blueprint(container, options) {
         };
 
         relationLabelsEntered.insert('text') // Add text layer of '&#9608;'s to erase the area on the path for label text.
-            .attr('class', _configHolder.nodes.relationship.labelSelector)
+            .attr('class', _configHolder.nodes.relationship.labelClass)
             .attr('dominant-baseline', 'middle')
             .attr('text-anchor', 'middle')
             .attr('font-family', 'monospace')
@@ -908,7 +908,7 @@ export function D3Blueprint(container, options) {
             .html((d) => ('&#9608;'.repeat(getLabelText(d).length + 2))); // +2 spaces for padding
 
         relationLabelsEntered.insert('text') // Add label text on top of '&#9608;'s which is on top of the path.
-            .attr('class', _configHolder.nodes.relationship.labelSelector)
+            .attr('class', _configHolder.nodes.relationship.labelClass)
             .attr('dominant-baseline', 'middle')
             .attr('text-anchor', 'middle')
             .attr('font-family', 'monospace')
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -782,15 +782,15 @@ export function D3Blueprint(container, options) {
         _d3DataHolder.visible.relationships.forEach(relation => {
             const directionKey = getRelationId(relation);
             if (!relationshipsPerDirection.has(directionKey)) {
-                relationshipsPerDirection.set(directionKey, []);
+                relationshipsPerDirection.set(directionKey, new Set());
             }
-            relationshipsPerDirection.get(directionKey).push(relation.label);
+            relationshipsPerDirection.get(directionKey).add(relation.label);
         });
 
         // Calculates offset for label, based on number of labels in the same direction.
         const getLabelOffset = (relationDataItem) => {
             const directionKey = getRelationId(relationDataItem);
-            let labelIndex = relationshipsPerDirection.get(directionKey).findIndex(label => label === relationDataItem.label);
+            let labelIndex = Array.from(relationshipsPerDirection.get(directionKey)).indexOf(relationDataItem.label);
             return labelIndex > 0 ? labelIndex * -12 : 0;
         };
 
--- a/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
+++ b/ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
@@ -774,18 +774,38 @@ export function D3Blueprint(container, options) {
     function drawRelationships() {
         showRelationships();
 
+        // Generates unique relation ID string.
+        const getRelationId = (relationDataItem) => (relationDataItem.source._id + '_related_to_' + relationDataItem.target._id);
+
+        // Group relationships per direction, as a key.
+        let relationshipsPerDirection = new Map();
+        _d3DataHolder.visible.relationships.forEach(relation => {
+            const directionKey = getRelationId(relation);
+            if (!relationshipsPerDirection.has(directionKey)) {
+                relationshipsPerDirection.set(directionKey, []);
+            }
+            relationshipsPerDirection.get(directionKey).push(relation.label);
+        });
+
+        // Calculates offset for label, based on number of labels in the same direction.
+        const getLabelOffset = (relationDataItem) => {
+            const directionKey = getRelationId(relationDataItem);
+            let labelIndex = relationshipsPerDirection.get(directionKey).findIndex(label => label === relationDataItem.label);
+            return labelIndex > 0 ? labelIndex * -12 : 0;
+        };
+
         let relationData = _relationGroup.selectAll('.relation')
-            .data(_d3DataHolder.visible.relationships, (d)=>(d.source._id + '_related_to_' + d.target._id));
+            .data(_d3DataHolder.visible.relationships, (d) => getRelationId(d));
 
         let relationDataEntered = relationData.enter();
 
         // Draw the relationship path
         relationDataEntered.insert('path')
             .attr('class', (d) => ('relation ' + d.pathSelector))
-            .attr('id', (d)=>(d.source._id + '-' + d.target._id))
+            .attr('id', (d) => getRelationId(d))
             .attr('opacity', 0)
-            .attr('from', (d)=>(d.source._id))
-            .attr('to', (d)=>(d.target._id));
+            .attr('from', (d) => (d.source._id))
+            .attr('to', (d) => (d.target._id));
 
         // Draw the relationship label that follows the path, somewhere in the middle.
         // NOTE `textPath` DECREASES THE UI PERFORMANCE, USE LABELS WITH CAUTION.
@@ -797,9 +817,11 @@ export function D3Blueprint(container, options) {
             .attr('font-family', 'monospace')
             .attr('fill', '#f5f6fa') // colour of the canvas
                 .insert('textPath')
-                .attr('xlink:href', (d)=>('#' + d.source._id + '-' + d.target._id))
+                .attr('xlink:href', (d) => ('#' + getRelationId(d)))
                 .attr('startOffset', '59%') // 59% roughly reflects `middle of the arch` minus `node radius`.
-                .html((d) => ('&#9608;'.repeat(d.label.length + 2)));
+                    .insert('tspan')
+                    .attr('dy', (d) => getLabelOffset(d))
+                    .html((d) => ('&#9608;'.repeat(d.label.length + 2)));
         relationDataLabelsEntered.insert('text') // Add label text on top of '&#9608;'s which is on top of the path.
             .attr('class', (d) => (d.labelSelector))
             .attr('dominant-baseline', 'middle')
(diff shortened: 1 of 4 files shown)


tbouron:
@algairim What happens if you have, say 20 relationships? If I'm reading the code correctly, it would display all the 20 relationship names right?

It is not very likely to happen but since #241, there is a high potential to have quite a lot of relationships. I think we should take that into account and limit the maximum number of label appearing there.

What about displaying up to 3 labels (3 is arbitrary, could be more or less by default) and if there are more relationships, display
```
relationship-1
relationship-2
relationship-3
+ 2 others
```
Hovering over the label would display all the labels.

algairim:
> @algairim What happens if you have, say 20 relationships? If I'm reading the code correctly, it would display all the 20 relationship names right?
> 
> It is not very likely to happen but since #241, there is a high potential to have quite a lot of relationships. I think we should take that into account and limit the maximum number of label appearing there.
> 
> What about displaying up to 3 labels (3 is arbitrary, could be more or less by default) and if there are more relationships, display
> 
> ```
> relationship-1
> relationship-2
> relationship-3
> + 2 others
> ```
> 
> Hovering over the label would display all the labels.

@tbouron, please note, these are not angular compiled components. Suggestions how to achieve such behaviour without angular are welcome.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -774,18 +774,38 @@ export function D3Blueprint(container, options) {
     function drawRelationships() {
         showRelationships();
 
+        // Generates unique relation ID string.
+        const getRelationId = (relationDataItem) => (relationDataItem.source._id + '_related_to_' + relationDataItem.target._id);
+
+        // Group relationships per direction, as a key.
+        let relationshipsPerDirection = new Map();
```
jathanasiou, line 781: Do we need strict order iteration for `relationshipsPerDirection`? If not, it might be simpler to use a POJO `{}` instead.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -774,18 +774,38 @@ export function D3Blueprint(container, options) {
...
+            if (!relationshipsPerDirection.has(directionKey)) {
+                relationshipsPerDirection.set(directionKey, []);
+            }
+            relationshipsPerDirection.get(directionKey).push(relation.label);
+        });
+
+        // Calculates offset for label, based on number of labels in the same direction.
+        const getLabelOffset = (relationDataItem) => {
+            const directionKey = getRelationId(relationDataItem);
+            let labelIndex = relationshipsPerDirection.get(directionKey).findIndex(label => label === relationDataItem.label);
```
jathanasiou, line 793: Since the arrays have string elements we can simply used `.indexOf` here like
`let labelIndex = relationshipsPerDirection.get(directionKey).indexOf(relationDataItem.label);`

algairim: Improved in 8fa7e4decb16d1a1abd7eb93bd0bae088f44023b

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Agree, can be a POJO instead of Map. Applied in 49f0dff9a3eab634d497228a6a4b339e7583eb22.

tbouron:
> @tbouron, please note, these are not angular compiled components. Suggestions on how to achieve such behaviour without angular are very welcome.

@algairim It would be back to the basic of DOM `mouseover` event to display a particular `div`. See this example: https://bl.ocks.org/d3noob/a22c42db65eb00d4e369

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -838,11 +827,77 @@ export function D3Blueprint(container, options) {
...
             .transition()
             .duration(_configHolder.transition)
             .attr('opacity', 0)
             .remove();
+
+        // ====== RELATIONSHIP LABELS =========
+        // Draw relationship labels that follow paths, somewhere in the middle of paths.
+        // NOTE <textPath/> DECREASES THE UI PERFORMANCE, USE LABELS WITH CAUTION.
+
+        _relationLabels.selectAll('.relation-label').remove(); // Re-draw labels every time, required to refresh changes.
```
algairim, line 840: Re-binding data for multiple labels in a single direction does not resolve the case when paths are re-assigned, especially when there can be more than one relationship type per direction. For that reason - re-draw them every time.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -838,11 +827,77 @@ export function D3Blueprint(container, options) {
...
+                labelSelector: d.labelSelector,
+                label: d.label
+            };
+            return accumulator;
+        }, {}));
+
+        let relationLabelsEntered = _relationLabels.selectAll('.relation-label').data(labelsData).enter();
+
+        // TODO: limit the number of labels that can be displayed for a path. Consider displaying up to 2-3 labels and
+        //       if there are more relationships in a single path - display '+ N others' at the end.
```
algairim, line 872: @tbouron, this is to address as per your suggestion. Either now or later - to agree with @ahgittin.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -771,50 +774,37 @@ export function D3Blueprint(container, options) {
...
 
-        let relationData = _relationGroup.selectAll('.relation')
-            .data(_d3DataHolder.visible.relationships, (d)=>(d.source._id + '_related_to_' + d.target._id));
+        const getPathId = (d) => (d.source._id + '-' + d.target._id);
+        const getRelationId = (d) => (getPathId(d) + '-' + d.pathSelector);
+
+        // ====== RELATIONSHIP ARCS ===========
 
-        let relationDataEntered = relationData.enter();
+        let arcsData = Object.values(_d3DataHolder.visible.relationships.reduce((accumulator, d) => {
```
jathanasiou, line 789: Due to only pulling `Object.values`, in the end `arcsData` will be an array of all the `{source,target,pathSelector}` objects it seems. If so, it might be easier to just do

```
const arcsData = _d3DataHolder.visible.relationships;
// or in case it's important to retain specific properties only
const arcsData = _d3DataHolder.visible.relationships
  .map(({ source, target, pathSelector }) => ({ source, target, pathSelector }));
```

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -771,50 +774,37 @@ export function D3Blueprint(container, options) {
...
+    /**
+     * Draws new relationships as arcs and labels. Updates existing ones.
+     */
     function drawRelationships() {
+
         showRelationships();
 
-        let relationData = _relationGroup.selectAll('.relation')
-            .data(_d3DataHolder.visible.relationships, (d)=>(d.source._id + '_related_to_' + d.target._id));
+        const getPathId = (d) => (d.source._id + '-' + d.target._id);
```
jathanasiou, line 784: Both arrow funcs _might_ be more readable like
```
const getPathId = ({ source, target }) => `${source._id}-${target._id}`;
```

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -771,50 +774,37 @@ export function D3Blueprint(container, options) {
...
+                pathSelector: d.pathSelector
+            };
+            return accumulator;
+        }, {}));
 
-        // Draw the relationship path
-        relationDataEntered.insert('path')
+        let relationArcs = _relationArcs.selectAll('.relation').data(arcsData, (d) => getRelationId(d));
+
+        relationArcs.enter().insert('path')
```
jathanasiou, line 800: Are we inserting a list of `<path>` elements to the DOM here?

If so, it would probably be more performant (and angular-like) to have the template html iterate something like `$scope.relationships` to generate them.

A `$watch` statement could keep that scope variable updated whenever any of the dependant properties for the `drawRelationships` logic changes.

algairim: @jathanasiou, this is not angular, this is D3, not sure if it applies.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: There is a need to reduce, not to map. See code comments.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Applied in 51765b60f9a72c4baf0db8bf57a3fd877f217300.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -771,50 +774,37 @@ export function D3Blueprint(container, options) {
...
-        let relationData = _relationGroup.selectAll('.relation')
-            .data(_d3DataHolder.visible.relationships, (d)=>(d.source._id + '_related_to_' + d.target._id));
+        const getPathId = (d) => (d.source._id + '-' + d.target._id);
+        const getRelationId = (d) => (getPathId(d) + '-' + d.pathSelector);
+
+        // ====== RELATIONSHIP ARCS ===========
 
-        let relationDataEntered = relationData.enter();
+        let arcsData = Object.values(_d3DataHolder.visible.relationships.reduce((accumulator, d) => {
+            accumulator[getRelationId(d)] = {
```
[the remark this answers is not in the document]
algairim: Agree, simplified in 51765b60f9a72c4baf0db8bf57a3fd877f217300.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
jathanasiou: Haven't touched D3 in a long time so not sure about how it interacts with the DOM. If D3 itself is calling `drawRelationships` and dealing with DOM updates, yeah it's probably not a good idea to bring angular templates into this.

algairim: It's just we will need to compile angular components on a fly to inject them here, which is not a good idea.

ahgittin:
my thoughts @algairim 

* if easy then if more than 3 relationships then show 2 and add a line which says "+2 relationships"
* if easy add a div `title` which should show a tooltip -- of maybe mouseover -- but def not angular (agree we don't want to compile/insert angular here)
* what happens if "requirements" and "relationships" are hidden in the layers menu -- are we able to filter?  if not we could combine the two into "relationships" and show/hide requirements and relationships together

algairim:
> my thoughts @algairim
> 
> * if easy then if more than 3 relationships then show 2 and add a line which says "+2 relationships"

Downstream changes complicate this, more than one type of relationships can be present at the same path, having different filters.

> * if easy add a div `title` which should show a tooltip -- of maybe mouseover -- but def not angular (agree we don't want to compile/insert angular here)
> * what happens if "requirements" and "relationships" are hidden in the layers menu -- are we able to filter?  if not we could combine the two into "relationships" and show/hide requirements and relationships together

We are able to filter "requirements" and "relationships" separately due to path/label selectors introduced recently. Combining the two into "relationships" and show/hide requirements and relationships together was the original behaviour.

algairim:
Added a tooltip for other labels if more than one exist per path.
Screenshot 2021-07-22 at 14 55 28

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -837,45 +850,50 @@ export function D3Blueprint(container, options) {
...
-            labelsPerPath[key].add(d.label);
+            relationshipsPerPath[key].add(getRelationId(d));
         });
 
+        const OFFSET_RATIO_PX = -12;
         const getLabelOffset = (d) => {
-            let labelIndex = Array.from(labelsPerPath[getPathId(d)]).indexOf(d.label);
-            return labelIndex > 0 ? labelIndex * -12 : 0;
+            let labelIndex = Array.from(relationshipsPerPath[getPathId(d)]).indexOf(getRelationId(d));
+            return labelIndex > 0 ? labelIndex * OFFSET_RATIO_PX : 0;
```
jathanasiou, line 866: Can be like so since the Set/array only contains string values.
```
  const hasLabel = Array.from(relationshipsPerPath[getPathId(d)]).includes(getRelationId(d));
  return hasLabel ? labelIndex * OFFSET_RATIO_PX : 0;
```

algairim: `labelIndex` is required to calculate the offset.

jathanasiou: Good point, I can't read.

algairim:
Agreed with @tbouron and @ahgittin to:
* split relationships layer into relationships arcs and labels, thus separate selectors;
* do not support different relationship types, as any association between nodes is a relationship after all.
* improve tooltip to follow the mouse and follow the panel style.

algairim:
This is how tooltip is displayed now:

Screenshot 2021-07-23 at 09 45 51

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -771,50 +781,35 @@ export function D3Blueprint(container, options) {
...
+     */
     function drawRelationships() {
+
         showRelationships();
 
-        let relationData = _relationGroup.selectAll('.relation')
-            .data(_d3DataHolder.visible.relationships, (d)=>(d.source._id + '_related_to_' + d.target._id));
+        const getPathId = ({ source, target }) => `${source._id}-${target._id}`;
+        const PATH_SELECTOR = 'relation-arc';
+        const LABEL_SELECTOR = 'relation-label';
```
tbouron, line 793: [MINOR] The convention up until now is to put constant into the `_configHolder`

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -838,11 +833,104 @@ export function D3Blueprint(container, options) {
...
+        _d3DataHolder.visible.relationships.forEach(d => {
+            const key = getPathId(d);
+            if (!labelsPerPath[key]) {
+                labelsPerPath[key] = new Set();
+            }
+            labelsPerPath[key].add(d.label);
+        });
+
+        const getLabelId = (d) => (getPathId(d) + '-' + d.label);
+        const MAX_LABELS_TO_DISPLAY = 2;
```
tbouron, line 859: Same as above

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -838,11 +833,104 @@ export function D3Blueprint(container, options) {
...
+            if (labelsCollectedForPath <= MAX_LABELS_TO_DISPLAY && !accumulator[labelKey]) {
+                accumulator[labelKey] = d;
+                accumulator[labelKey].labels = Array.from(labelsPerPath[pathKey]); // path labels.
+                accumulator[labelKey].labelIndex = labelsCollectedForPath; // label index at path.
+            }
+            return accumulator;
+        }, {});
+
+        const CANVAS_COLOR = '#f5f6fa';
+        const OFFSET_RATIO_PX = -12;
```
tbouron, line 875: Same as above

ui-modules/blueprint-composer/app/views/main/main.controller.js
```diff
@@ -110,8 +117,37 @@ export function MainController($scope, $element, $log, $state, $stateParams, brB
...
+            relationshipLabelsLayer.active = false;
+        }
+
+        // Update cached state for arc and labels.
+        isRelationshipArcsLayerActive = relationshipArcsLayer.active;
+        isRelationshipLabelsLayerActive = relationshipLabelsLayer.active;
+
     }, true);
 
+    $scope.$on('layers.filter', () => applyFilters());
```
tbouron, line 149: Why do we need to apply the filter when the graph is redrawn? Sounds like the logic is backward. The entrypoint is the click on the filters, isn't it?

tbouron (changes requested):
Looks good @algairim. I just have few minor comments + one question, see below

ui-modules/blueprint-composer/app/views/main/main.controller.js
algairim: We need to redraw it for cases when filter is changed first and graph is changed after. Filter change in Layers applies changes styling changes to all existing elements on the canvas. Re-draw event is to re-apply those styling effects for graph changes.

tbouron: Aha, I see @algairim. That is a bit counter-intuitive but probably the best solution for now (and I wouldnt want to make you change even more stuff now :P)

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Addressed in 94485d2c28c48a6aadb92f167fddc284eab02e8f.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Addressed in 94485d2c28c48a6aadb92f167fddc284eab02e8f.

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
algairim: Addressed in 94485d2c28c48a6aadb92f167fddc284eab02e8f.

tbouron (approved):
LGTM, thanks @algairim 👍 
I'll merge once Jenkins is happy

PR #256, opened 2021-08-02
Display node ID in the Scope dropdown menu of the DSL editor

algairim:

### DIFF
--- a/ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
+++ b/ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
@@ -89,15 +89,11 @@ export function dslEditorDirective($rootScope, $filter, $log, brUtilsGeneral, bl
                 attrs.push('parent');
             }
 
-            let name = item.name;
-            if (attrs.length > 0) {
-                name += ` (${attrs.join(', ')})`;
-            } else {
-                name += ` (${item.entity.id || item.id})`;
-            }
+            let { name, entity, id } = item;
+            name += attrs.length > 0 ? ` (${attrs.join(', ')})` : ` (${entity.id || id})`;
 
             return {
-                id: item.id,
+                id,
                 label: name,
                 scope: 'On specific entity'
             };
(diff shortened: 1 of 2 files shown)


algairim:
Proposed improvement to distinguish nodes in the Scope dropdown menu:
Screenshot 2021-08-02 at 13 45 18

ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
```diff
@@ -92,6 +92,8 @@ export function dslEditorDirective($rootScope, $filter, $log, brUtilsGeneral, bl
             let name = item.name;
             if (attrs.length > 0) {
                 name += ` (${attrs.join(', ')})`;
+            } else {
```
jathanasiou, line 95: Might be cleaner as

```
let { name, entity, id } = item;
name += (attrs.length > 0)
  ? ` (${attrs.join(', ')})`
  : ` (${entity.id || id})`;
```

jcabrerizo (approved):
Looks good to me @algairim

ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
jathanasiou: Addendum: if this is used, remember to adjust the return statement as well.
```
    return {
        id,
        label: name,
        scope: 'On specific entity',
    };
```

algairim: Good suggestion, I'll take it if more changes will be required in this PR.

ahgittin:
@algairim should we also improve the entity identifying info in the blocks underneath the dropdown (lower right, where it says just Basic Entity) ?

image

in addition i wonder if it's not too hard and you think it would be useful -- could we intercept clicks on entities in the canvas when the DSL editor is open and use those clicks to change the DSL editor scope to the entity selected on canvas?

algairim:
@ahgittin, we can display ID for every item in the search result, like for `Basic Entity (ID: e1)`, at the moment it is displayed for `item.type === DSL_KINDS.ENTITY` only as on the screenshot.
As for changing the DSL editor scope: click on the canvas destroys right-hand-side menu and draws a new one for the node clicked. It will cost some effort to re-design that behaviour.

ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
algairim: Improved in a6899562a7a127dac99f64040f90f8441bc7d305.

algairim:
IDs are now displayed for every item in the search result.

ahgittin:
:+1:

PR #257, opened 2021-08-03
Fix/dsl editor duplicates

jathanasiou:
Cleaned up the DSL Editor entry generation logic and added a uniqueness filter based on the `id` property to avoid duplicates.


ahgittin:
nice one, merging

PR #261, opened 2021-08-04
Fixed Composer required label check

jathanasiou:
Default values were not being checked properly from Parameters and Configuration fields erroneously had *required* label


iuliana (approved):
LGTM

iuliana:
The test failure is not related to the changes. I will merge this, because I am pretty confident  it won't break anything, but move things forward on other issues being worked on.

ahgittin:
aha this is part of the fix -- `parameter` uses `default` -- however _config_ key definitions use `defaultValue` so this will break for defaults coming from the server-side model (eg parent class).  and there are other places we use `defaultValue` where we should be checking both `defaultValue` and `default`.  will open follow-on PR.

PR #262, opened 2021-08-04
Use 'component' instead of 'parent' when sugesting parent property va…

algairim:
…lue as a quick-fix


jathanasiou (approved):
Looks good.

PR #263, opened 2021-08-05
Parameter defaults

ahgittin:
Broader fix for issue observed in #261

### DIFF
--- a/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
+++ b/ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
@@ -323,7 +323,7 @@
           <div ng-if="model.miscData.get('config').length === 0">
             <h4>No configuration</h4>
             <p class="buttons">
-                <button class="btn btn-sm btn-success" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="state.config.search">
+                <button class="btn btn-sm btn-success" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="state.config.search && state.config.search!=''">
                         <i class="fa fa-plus"></i> Add '{{state.config.search}}'
                 </button>
             </p>
@@ -332,7 +332,7 @@
             <h4>No matching configuration</h4>
             <p class="buttons">
                 <button class="btn btn-sm btn-default" ng-if="state.config.search.length > 0" ng-click="state.config.search = ''">Clear search</button>
-                <button class="btn btn-sm btn-default" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="!specEditor.getConfig(state.config.search)">
+                <button class="btn btn-sm btn-default" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="state.config.search && state.config.search!='' && !specEditor.getConfig(state.config.search)">
                         <i class="fa fa-plus"></i> Add '{{state.config.search}}'
                 </button>
                 <button class="btn btn-sm btn-success" ng-if="!state.config.filter.values.all" ng-click="state.config.filter.values.all = true">
--- a/ui-modules/blueprint-composer/app/components/util/model/entity.model.js
+++ b/ui-modules/blueprint-composer/app/components/util/model/entity.model.js
@@ -685,7 +685,17 @@ function addConfigKeyDefinition(param, overwrite, skipUpdatesDuringBatch) {
         let key = (param || {}).name;
         if (!key) throw new Error("'name' field must be included when adding parameter; was", param);
 
-        allConfig[key] = Object.assign(allConfig[key] || {}, param, overwrite ? null : allConfig[key]);
+        let paramMapped = Object.assign({}, param);
+        let configDef = allConfig[key] || {};
+        if (typeof paramMapped.default !== 'undefined') {
+            /* Annoyingly, in parameters, we call the default value "default",
+             * but in config, we call them "defaultValue";
+             * when params are merged to config we need to rename
+             */
+            paramMapped.defaultValue = paramMapped.default;
+            delete paramMapped['default'];
+        }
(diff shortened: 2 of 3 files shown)


ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
```diff
@@ -323,7 +323,7 @@ <h4>No matching parameters</h4>
           <div ng-if="model.miscData.get('config').length === 0">
             <h4>No configuration</h4>
             <p class="buttons">
-                <button class="btn btn-sm btn-success" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="state.config.search">
+                <button class="btn btn-sm btn-success" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="state.config.search && state.config.search!=''">
```
jathanasiou, line 326: Don't believe this is needed as it adds complexity and  empty `''` strings evaluate to false in `if` statements anyways. https://stackoverflow.com/questions/27380000/angularjs-ng-if-check-string-empty-value

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
```diff
@@ -332,7 +332,7 @@ <h4>No configuration</h4>
             <h4>No matching configuration</h4>
             <p class="buttons">
                 <button class="btn btn-sm btn-default" ng-if="state.config.search.length > 0" ng-click="state.config.search = ''">Clear search</button>
-                <button class="btn btn-sm btn-default" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="!specEditor.getConfig(state.config.search)">
+                <button class="btn btn-sm btn-default" ng-click="specEditor.addConfigKey(state.config.search)" ng-if="state.config.search && state.config.search!='' && !specEditor.getConfig(state.config.search)">
```
jathanasiou, line 335: Same as previous.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
```diff
@@ -685,7 +685,17 @@ function addConfigKeyDefinition(param, overwrite, skipUpdatesDuringBatch) {
         let key = (param || {}).name;
         if (!key) throw new Error("'name' field must be included when adding parameter; was", param);
 
-        allConfig[key] = Object.assign(allConfig[key] || {}, param, overwrite ? null : allConfig[key]);
+        let paramMapped = Object.assign({}, param);
```
jathanasiou, line 688: I feel confident the below would be a much cleaner version of this function with the same behaviour. I'd really suggest something close for maintenance purposes if it feels accurate logic-wise.

```javascript
function addConfigKeyDefinition(param, overwrite, skipUpdatesDuringBatch) {
    const allConfig = this.miscDataOrDefault('configMap', {});
    
    if (typeof param !== 'undefined') {
        
        const parameterItem = (typeof param !== 'string')
            ? { ...param } // shallow copy as object
            : {
                "name": param,
                "label": param,
                "description": "",
                "priority": 1,
                "pinned": true,
                "type": "java.lang.String",
                "constraints": [],
            };

        const key = get(parameterItem, 'name'); // import { get } from 'lodash' at top, defaults to undef.

        // this will also throw if original `param` is neither object nor string due to previous statement
        if (!key) throw new Error("'name' field must be included when adding parameter; was", param);

        if (typeof parameterItem.default !== 'undefined') {
            /* Annoyingly, in parameters, we call the default value "default",
             * but in config, we call them "defaultValue";
             * when params are merged to config we need to rename
             */
            parameterItem.defaultValue = parameterItem.default;
            delete parameterItem.default;
        }

        Object.assign(allConfig[key], parameterItem);
    }

    if (!skipUpdatesDuringBatch) {
        this.miscData.set('config', Object.values(allConfig));
    }

    this.touch();
    return this;
}
```

jathanasiou (changes requested):

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
ahgittin: that's what i thought @jathanasiou but for some reason without this i was seeing an `Add ''` button and adding this seemed to fix it!  not sure what is going on.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
ahgittin: slightly clearer and more efficient apart from (1) you've lost the `overwrite` behaviour and (2) bringing in `get` obscures behaviour for those not familiar with `_.get` -- but the benefit is so slight it isn't worth the effort of testing it

ui-modules/blueprint-composer/app/components/spec-editor/spec-editor.template.html
jathanasiou: Sounds like an angular 1 bug. I saw some mentions online of how `ng-if` evaluates expressions so maybe we're victims of that.

ui-modules/blueprint-composer/app/components/util/model/entity.model.js
jathanasiou: Yeah fair, just though I'd share. For the record, `overwrite` was removed because it seemed to contribute no different logic path to the end result due to the syntax I went with that allowed a simpler extension.

ahgittin: if `overwrite` was false or set false because `param` was supplied as a string, then the difference is that the code should do `Object.assign(parameterItem, allConfig[key])` prior to the `Object.assign(allConfig[key], parameterItem)` -- the difference being if both set a value for the same key, which one should be kept.  With this change any key set in the param object will always overwrite such a key in config.

ahgittin:
thx for the review @jathanasiou -- merging

PR #264, opened 2021-08-06
Reference scopeRoot in DSL instead of component by ID if referenced n…

algairim:
…ode is a root


ahgittin:
LGTM!

PR #267, opened 2021-08-16
adjusted logic for identifying unique items

jathanasiou:


ui-modules/blueprint-composer/app/components/dsl-editor/dsl-editor.js
```diff
@@ -305,9 +305,10 @@ export function dslEditorDirective($rootScope, $filter, $log, brUtilsGeneral, bl
     function uniqueItems(items) {
         const IDs = new Set();
 
-        return items.filter(({ id }) => {
-            if (IDs.has(id)) return false;
-            IDs.add(id);
+        // filtering with both own and parent's ID in case we have same-type child nodes
+        return items.filter(({ id, entity }) => {
```
ahgittin, line 309: makes sense.  for readability and in case entity.id is null and for clean demarcation between id and entity.id i'd do this as:

```
let marker = id + ':' + (entity.id || '-');
if (IDs.has(marker)) {
  return false;
} else {
  IDs.add(marker);
  return true;
}
```

PR #272, opened 2021-08-30
Entity sorting correction

jathanasiou:
Added a default displayName=symbolicName to prevent unwanted clustering

Avoids fallback to 'lastUsed' attribute with which elements are moved around as they are used


algairim:
Are changes in `rest-api-docs` part of this PR?

jathanasiou:
> 
> Are changes in `rest-api-docs` part of this PR?

Nope, my bad will adjust the PR now.

PR #273, opened 2021-08-30
Entity sorting correction

jathanasiou:
Added a default displayName=symbolicName to prevent unwanted clustering

Avoids fallback to 'lastUsed' attribute with which elements are moved around as they are used


algairim (approved):
Looks good.

PR #276, opened 2021-09-01
Persistence import directive / export download implementation

zan-mateusz:
Introducing a directive to allow uploading (importing) of a persistence store archive (currently zip only).
Introducing capability to download (export) the current persistent store.


iuliana (approved):
Looks good to me.

PR #287, opened 2021-09-15
UI awareness of configurable sensitive fields

ahgittin:
sensitive fields and whether deployments are blocked are now loaded from the server and set dynamically where used (eg in config values table); and in composer, if plaintext values are blocked a warning appears when the field is being edited

image

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -22,6 +22,7 @@ import {Issue, ISSUE_LEVEL} from '../util/model/issue.model';
 import {Dsl} from "../util/model/dsl.model";
 import jsYaml from "js-yaml";
 import typeNotFoundIcon from "../../img/icon-not-found.svg";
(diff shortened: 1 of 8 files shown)


ahgittin:
depends on https://github.com/apache/brooklyn-server/pull/1258 of course

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -427,6 +428,31 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
     }
 
     function refreshConfigConstraints(entity) {
+        function checkSensitiveFields(config) {
+            console.log("checkConstraints", config);
```
algairim, line 432: Debugging log?

algairim (approved):

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
ahgittin: good spot, fixed, force-pushed

PR #288, opened 2021-09-15
ensuring that tosca post processors are not applied if type is camp

zan-mateusz:
For support of both camp and tosca types at the same time:

Previously:
camp type had a post processor applied to it, that translated it to tosca
API to process the deployment on the server side received the tosca 1.3 template with format as tosca13
server translate the template back to camp and deploys the application

Above is unnecessary conversion back and forward and also has some side effects (such as an unnecessary delete effector).

With the change:
if using camp type, we don't apply post processors
API sends a camp template to the server

### DIFF
--- a/ui-modules/utils/quick-launch/quick-launch.js
+++ b/ui-modules/utils/quick-launch/quick-launch.js
@@ -150,8 +150,7 @@ export function quickLaunchDirective() {
 
         function deployApp() {
             $scope.deploying = true;
-            let useExtensions = $scope.app.plan.format.includes('camp') ? false : true;
-            Promise.resolve(quickLaunch.buildYaml(useExtensions))
+            Promise.resolve(quickLaunch.buildYaml($scope.app.plan.format.includes('camp') ? false : true))
                 .then(appYaml => {
                     quickLaunch.planSender(appYaml)
                         .then((response) => {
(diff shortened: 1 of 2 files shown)


ui-modules/utils/quick-launch/quick-launch.js
```diff
@@ -150,8 +150,8 @@ export function quickLaunchDirective() {
 
         function deployApp() {
             $scope.deploying = true;
-
-            Promise.resolve(quickLaunch.buildYaml())
+            let useExtensions = $scope.app.plan.format.includes('camp') ? false : true;
+            Promise.resolve(quickLaunch.buildYaml(useExtensions))
```
jathanasiou, line 154: `includes` returns boolean so much less redundant to simply do:

Promise.resolve(quickLaunch.buildYaml(!$scope.app.plan.format.includes('camp')))

Pretty good otherwise

jathanasiou (approved):

iuliana (approved):
LGTM, just make sure you test the project using only the .. ahem ... extensions.

ui-modules/utils/quick-launch/quick-launch.js
zan-mateusz: addressed the comment. Thanks

PR #290, opened 2021-09-16
Enhancement/sensitive field warning icon - regression fix

jathanasiou:


ui-modules/app-inspector/app/views/main/inspect/summary/summary.controller.js
```diff
@@ -100,15 +101,10 @@ export function summaryController($scope, $state, $stateParams, $q, $http, $http
...
-                        key,
-                        value,
-                        // marking as unsafe if the field name looks sensitive
-                        // and the unresolved value does *not* come from a secure external source
-                        isUnsafe: isSensitiveFieldName(key.trim()) &&
-                            !vm.config[key].toString().startsWith('$brooklyn:'),
-                    }));
+                vm.configItems = vm.showResolvedConfig ? vm.configResolved : vm.config;
+                vm.configItemsUnsafeMap = mapValues(vm.configItems, (value, key) =>
+                    isSensitiveFieldName(key.trim()) && !vm.config[key].toString().startsWith('$brooklyn:')
```
ahgittin, line 106: will this fail if not set?

i will test and fix

ahgittin: ie `vm.config[key]` can be null can't it?

also do we need `mapValues` ... can't we just use Object.keys(vm.configItems).forEach(key => ...)`?

finally on reflection i think it would be cleaner to pass a function eg `isSensitive(key, value)` to the widget, rather than a map.

if no objections i'll apply this and merge.

jathanasiou: the lodash `mapValues` (like most of its functions) is very resistant to "bad" values, so worse scenario we'll have an empty object for the config-sensor table, meaning no warning icons at all but everything else should be fine.

image

ahgittin: i knew i'd learn something by asking!  ta

ahgittin:
tidy applied in ed9096273120f3d17f260b24c86fceb1f125a83c, and all merged 👍

PR #291, opened 2021-09-17
Fix/home page quick launch

zan-mateusz:
Setting the format value correctly when deploying from home page quick launch


iuliana (approved):
LGTM

PR #293, opened 2021-10-01
use new deleteable flag on bundles to give better info on delete

ahgittin:
This adds popups on catalog-item deletion (bundle and locations; no delete for type, though some support is in here for that) which:
* advise if you are are deleting a non-deleteable bundle, that the deletion is temporary
* if you are deleting a type from a bundle (only a location supported through UI at present), it advises whether the bundle will also be deleted, and if not it gives you the option to review it and delete it instead
* advises that if you have deployments using the type bad things can happen

For example:

location in dedicated single-location bundle

image

and location in larger bundle:

image

and user bundle:

image

and system bundle:

image

### DIFF
--- a/ui-modules/utils/catalog-deleter/catalog-deleter.js
+++ b/ui-modules/utils/catalog-deleter/catalog-deleter.js
@@ -1,0 +1,145 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import angular from 'angular';
+import template from './catalog-deleter.html';
+import catalogApi from '../providers/catalog-api.provider';
+
+const MODULE_NAME = 'brooklyn.components.catalog-deleter';
+
+/**
+ * @ngdoc module
+ * @name brooklyn.components.catalog-deleter
+ * @requires catalogApi
+ *
+ * @description
+ * Adds an overlay on top of the current DOM element to upload files to the catalog. Files can either by added via
+ * classic file selection or drag & drop. This support multiple files to be uploaded at once.
+ */
+angular.module(MODULE_NAME, [catalogApi])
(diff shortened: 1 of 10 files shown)


ahgittin:
Depends on https://github.com/apache/brooklyn-server/pull/1259

ui-modules/utils/catalog-deleter/catalog-deleter.less
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+.brooklyn-catalog-deleter {
+}
```
algairim, line 20: Unused.

ui-modules/utils/catalog-deleter/catalog-deleter.js
```diff
@@ -0,0 +1,145 @@
...
+            if ($scope.onDeleting) $scope.onDeleting();
+            let promise;
+            if ($scope.mode==='bundle') {
+                promise = catalogApi.deleteBundle($scope.symbolicName, $scope.version)
+            } else if ($scope.mode==='location') {
+                promise = catalogApi.deleteLocation($scope.symbolicName, $scope.version)
+            } else if ($scope.mode==='type') {
+                // not used
+                throw 'deleteType not supported';
+            } else {
```
algairim, line 120: How about restricting deletion with `ng-if`, so that button is available/visible for locations and bundles?

ui-modules/utils/catalog-deleter/catalog-deleter.js
```diff
@@ -0,0 +1,145 @@
...
+                $scope.bundleLoading = false;
+            });
+        }
+
+        if ($scope.mode==='bundle') {
+            $scope.bundleLoading = true;
+            getBundle($scope.symbolicName, $scope.version);
+
+        } else {
+            $scope.bundleLoading = true;
```
algairim, line 78: `$scope.bundleLoading = true;` can be moved up, away from `if` and `else`.

algairim (approved):
Looks good, couple of minor comments.

ui-modules/utils/catalog-deleter/catalog-deleter.less
ahgittin: i kinda like to have empty css rules for classes on directives -- even if they don't have any style yet there is a clear place it can be added.  i wouldn't go out of my way to make it but in this case i copied `updater` and so it was already there, seemed better to keep it empty than remove it.

ui-modules/utils/catalog-deleter/catalog-deleter.js
ahgittin: the caller controls the button, this simply control the popup, which i think is right for maximum reusability

ui-modules/utils/catalog-deleter/catalog-deleter.js
ahgittin: because there is corresponding "finally" logic in each block which needs to clear this, personally i think it is easier to follow with the duplication.

PR #299, opened 2021-10-20
Fix UI issues

duncangrant:
This PR fixes some issues with the UI:
Context stays on deleted app on canvas
Unclear error when blueprint has space in name
No visual feedback when unauthorised user tries to change HS state


ahgittin:
These have all been merged already.  PR can be closed.

PR #312, opened 2021-11-19
Do not restore search when the icon triggers selection directly or sh…

algairim:
…ows popup in catalog-selector


jathanasiou (approved):
Looks clean and straightforward, shouldn't have conflicts with current "search query caching".

PR #317, opened 2022-04-14
Update node and npm

jcabrerizo:


tbouron (approved):
LGTM @jcabrerizo. SInce Jenkins is happy, I think it is safe to merged

PR #322, opened 2022-05-06
Adjust width of switch view dropdown automatically

algairim:


iuliana (approved):

ahgittin:
sometimes `min-width` is nice in dropdowns but we can see later whether that is needed here or anywhere

PR #323, opened 2022-05-06
Do not handle Backspace key event

algairim:


iuliana (approved):

ui-modules/blueprint-composer/app/components/util/d3-blueprint.js
```diff
@@ -364,7 +364,9 @@ export function D3Blueprint(container, options) {
     function onKeyUp() {
         d3.event.stopPropagation();
```
ahgittin, line 365: Isn't the usual pattern to stop propagation only if the event is handled?

Happy to leave like this but one to watch if there are any related issues.

algairim: No part of this change-set.

PR #327, opened 2022-05-23
WIP: Added logic to check extra contraints

iuliana:
Added logic to check extra constraints and show error messages for them. 
1) Since the list of constraints is based on types that do not exist in Brooklyn, an impromevent of this would be to actually add those types in Brooklyn.
2) Transforming this bit of code into an extension and moving it to the dependent project looks like a pain to me, however if desired I am sure it is possible. 

Cheers!

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -43,6 +43,9 @@ export const COMMON_HINTS = {
     }]
 };
 
+// needed to support error messages for extra constraints
+export const EXTRA_CONSTRAINTS = ['equal', 'greater_than', 'greater_or_equal', 'less_than ', 'less_or_equal ', 'in_range ', 'valid_values', 'length ', 'min_length ', 'max_length ', 'pattern'];
+
 export function blueprintServiceProvider() {
     return {
         $get: ['$log', '$q', '$sce', 'paletteApi', 'iconGenerator', 'dslService', 'brBrandInfo',
@@ -481,12 +484,12 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
                         }
                         break;
                     case 'required':
-                        if (!isSet() && !hasDefault && val()!='') {
+                        if (!isSet() && !hasDefault && val()!=='') {
                             message = `<samp>${config.name}</samp> is required`;
                         }
                         break;
                     case 'regex':
-                        if (isSet() && !(new RegExp(args).test(val))) {
+                        if (isSet() && !(new RegExp(args).test(val()))) {
                             message = `<samp>${config.name}</samp> does not match the required format: <samp>${args}</samp>`;
                         }
                         break;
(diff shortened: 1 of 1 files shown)


ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -520,6 +523,70 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
...
+                                        message = `<samp>${config.name}</samp> must be in range of (inclusive) [${constrVals[0]} , ${constrVals[1]}]`;
+                                    }
+                                    break;
+                                case 'valid_values':
+                                    if (!isSet() || isSet() && constrVals[0].includes(val())) {
+                                        message = `<samp>${config.name}</samp> must be one of (${constrVals.toString()})`;
+                                    }
+                                    break;
+                                case 'length':
+                                    if (!isSet() || isSet() && val().length === constrVals[0]) {
```
geomacy, line 568: think this should be `val.length() !===`

geomacy (commented):
Looks good in general. A couple of comments.

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -520,6 +523,70 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
...
+                                        message = `<samp>${config.name}</samp> must be less than ${constrVals[0]}`;
+                                    }
+                                    break;
+                                case 'less_than_or_equal':
+                                    if (!isSet() || isSet() && val() > constrVals[0]) {
+                                        message = `<samp>${config.name}</samp> must be less than or equal to ${constrVals[0]}`;
+                                    }
+                                    break;
+                                case 'in_range':
+                                    if (!isSet() || isSet() && val() >= constrVals[0] && val() <= constrVals[1]) {
```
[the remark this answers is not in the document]
iuliana: Well depends how range is defined. For this particular scenario, the `in_range` is supposed to be inclusive.

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
iuliana: Yup, good catch! Thx!

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
```diff
@@ -43,6 +43,9 @@ export const COMMON_HINTS = {
     }]
 };
 
+// needed to support error messages for extra constraints
+export const EXTRA_CONSTRAINTS = ['equal', 'greater_than', 'greater_or_equal', 'less_than ', 'less_or_equal ', 'in_range ', 'valid_values', 'length ', 'min_length ', 'max_length ', 'pattern'];
```
algairim, line 47: Please add references to implementation of these extra constraints, either in the Brooklyn core or its dependencies.

algairim:
> Added logic to check extra constraints and show error messages for them.
> 
> 1. Since the list of constraints is based on types that do not exist in Brooklyn, an impromevent of this would be to actually add those types in Brooklyn.
> 2. Transforming this bit of code into an extension and moving it to the dependent project looks like a pain to me, however if desired I am sure it is possible.
> 
> Cheers!

We should consider moving implementation of constraints into Brooklyn core.

algairim:
We should consider moving implementation of constraints into Brooklyn core (did not mean to close this PR).

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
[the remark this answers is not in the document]
geomacy: Yeah but since we want to be inclusive to be valid, we shouldn't be using `>=` and `<=` to check for the invalid case. E.g. if `constrVals === [1, 10]` then 10 is valid (inclusive), but `val() <= constrVals[1]` will be `true` so the `if` will be true and you'll get the error message.

As this `if` is a quite a big chunk of JS, it would be nice to pull it out into a separate function and add some unit tests for it in the `.spec.js`.

ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
ahgittin: Also the trailing whitespace in some is odd.  Do we trim so it does not matter?

iuliana:
@geomacy Thank you for your input. We will be parking this until further notice.

PR #328, opened 2022-06-01
Fixes/catalog saver

algairim:


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
```diff
@@ -60,7 +60,7 @@ <h3 class="modal-title">{{title}}</h3>
                 <label class="control-label">Bundle ID</label>
                 <div class="input-group">
                     <span class="input-group-addon">catalog-bom-</span>
-                    <input ng-model="config.bundle" ng-disabled="state.saving" class="form-control" name="bundle" ng-pattern="state.pattern" autofocus placeholder="{{ defaultBundle || 'E.g. my-bundle-id' }}"/>
+                    <input ng-model="config.bundle" ng-disabled="isUpdate() || state.saving" class="form-control" name="bundle" ng-pattern="state.pattern" autofocus placeholder="{{ defaultBundle || 'E.g. my-bundle-id' }}"/>
```
algairim, line 63: @ahgittin, instead of suppressing duplicate version check if ID is changed, I suggest to block bundle ID change all together when template update launched from the catalog

Changing bundle ID on template update creates a mess with versions in the catalog, where a "new" bundle holds information about previous versions that do not exist this *new* bundle, and vice versa.

I did not see how such a mess is possible to create via CLI.

ui-modules/blueprint-composer/app/views/main/main.controller.js
```diff
@@ -148,10 +148,9 @@ export function MainController($scope, $element, $log, $state, $stateParams, brB
 
     $scope.$on('layers.filter', () => applyFilters());
 
-    vm.parseError = false;
```
algairim, line 151: This is not used anywhere.

ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.modal.template.html
ahgittin: sounds reasonable

PR #329, opened 2022-06-06
Fixed bug caused by object-based based constraints - no toUpperCase f…

iuliana:
Fixed bug caused by object-based based constraints - no toUpperCase function :D


jcabrerizo (approved):
LGTM @iuliana thanks

PR #325, opened 2022-05-17
Add `order` as module property and use it when is present before using name for ordering modules

jcabrerizo:
Allows to set up an order for listing the modules in the home page and the header menu. Also it renames the `App inspector` to just `Inspector` as it Apache Brooklyn can manage more than applications.

PR #333, opened 2022-06-09
Fix issue generated in 325

jcabrerizo:
Fix an UI error introduced by no injecting the `brUtilsGeneral` in all the modules in #325


algairim:
Works now, thanks for fixing this.

algairim:
Works now, thanks for fixing this.

PR #336, opened 2022-06-15
When an entity don't have a name, it shows the personalised entity `ID`

jcabrerizo:
When an entity don't have a name, it shows the entity `ID` only when is provided instead the `typeName`

If no ID is provided it still shows the `typeName` instead the interal generated id as is shown in the `id` field

https://user-images.githubusercontent.com/17095501/173815479-0b4f422a-7b9e-482d-a3ce-0e96b5d059b3.mov


iuliana (approved):
LGTM

PR #332, opened 2022-06-09
Refresh metadata for type Entity only

algairim:
Signed-off-by: Mykola Mandra <mykola.mandra@cloudsoft.io>

PR #338, opened 2022-06-23
Improve handling of spec types in composer

ahgittin:
Previously we used regex on key names to determine if the contents was an entity spec.  This meant some (eg multi-group) didn't work and in some cases could make the UI unusable or broken.

This checks the actual type defined for the key name and of the value if the type is Object and so supports entity spec

either if:
- it uses the DSL word
- the key type matches
- the key name matches (fallback for where type is unknown/not-loaded and value not set)

and in multiple places:
- in model
- in designer view canvas
- in spec editor

Also makes minor related improvements to UI and code.


ahgittin:
Includes and builds on https://github.com/apache/brooklyn-ui/pull/332

algairim (approved):

PR #334, opened 2022-06-10
Effector UX improvements

algairim:
Signed-off-by: Mykola Mandra <mykola.mandra@cloudsoft.io>

PR #339, opened 2022-06-23
Inspector -> Effector (& Activities) more restyling and more info and searchability

ahgittin:
builds on #334 to:
* put buttons at left (better use of estate)
* provide details of recent invocations (at right)
* allow search to be specified as query parameter `search` (so things can link to `.../effectors?search=my-effector`)

also supports specifying `search` and `filter` as query params to the activities view, so the link works, and actions there on effector tasks to reinvoke or view effectors

image


algairim (approved):
I tried these changes - great experience.

PR #340, opened 2022-06-24
don't hard-code a version in composer unless explicitly set

ahgittin:
versions set either in the yaml or by a user in the dropdown are fixed.
otherwise no version is stored.
fixes bad behaviour where previously, switching to composer would hard-code versions everywhere, _changing_ the blueprint (introduced in #292)

### DIFF
--- a/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
+++ b/ui-modules/blueprint-composer/app/components/providers/blueprint-service.provider.js
@@ -748,7 +748,7 @@ function BlueprintService($log, $q, $sce, paletteApi, iconGenerator, dslService,
             return result;
         }
         entity.clearIssues({group: 'type'});
-        entity.version = data.version;
+
         entity.type = data.symbolicName;
         entity.icon = entity.metadata.get('iconUrl')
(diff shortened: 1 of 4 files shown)


algairim:
This works nicely. Merging.

algairim (approved):

ahgittin:
Re-test this please.

algairim:
> Re-test this please.

Local build works - no test errors. And logic works. Jenkins is not happy.

ahgittin:
hmm all the UI tests seem to be failing, and giving the message

`DEBUG: 'Data in key nullKey is not a recognised entity spec; not treating as entity spec', TypeError{stack: 'addConfig@http://localhost:9876/base/karma.test.shim.js?1be23a675a6921f301e7668989a3dd2391c5d007:51304:19`

looks like this happened 3 days ago

some of the failures have

`java.io.IOException: Tried to load head FlowNodes for execution Owner[Brooklyn/brooklyn-ui/PR-338/2:Brooklyn/brooklyn-ui/PR-338 #2] but FlowNode was not found in storage for head id:FlowNodeId 1:40`

which is a spurious jenkins thing, so it's hard to be sure but naively it looks like https://github.com/apache/brooklyn-ui/commit/b41d064ecdea326852f09e4d055f2c7fabe95442 one line change is the culprit, if it's our code.  alternatively it could be something in the build environment.

if it passes locally, merge it, and if there are legitimate code problems or circleci problems, we'll resolve those separately

PR #345, opened 2022-07-11
Check if to drop stream filters for stream with CLI XML content only

algairim:

### DIFF
--- a/ui-modules/app-inspector/app/components/stream/stream.template.html
+++ b/ui-modules/app-inspector/app/components/stream/stream.template.html
@@ -19,10 +19,10 @@
 <loading-state error="error" ng-if="stream === undefined"></loading-state>
 
 <div ng-show="stream.length >= 0">
-  <pre ng-if="stream.length > 0 && !isFormatContent()" class="auto-scrollable">{{stream}}</pre>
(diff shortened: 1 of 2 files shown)


ui-modules/app-inspector/app/components/stream/stream.template.html
```diff
@@ -19,10 +19,10 @@
...
-  <pre ng-if="stream.length > 0 && isFormatContent() && !isFilterContent()" class="auto-scrollable">{{formattedStream}}</pre>
-  <pre ng-if="stream.length == 0" class="text-center"><i>(no content)</i></pre>
+  <pre ng-show="stream.length > 0 && !isFormatContent()" class="auto-scrollable">{{stream}}</pre>
+  <pre ng-show="stream.length > 0 && isFormatContent() && isFilterContent()" class="auto-scrollable"><div ng-if="isFilterContent()"><pre ng-repeat="item in filteredStream track by item.id" ng-show="isDisplayFormattedItem(item)" class="log-section" ng-class="getFormattedItemLogLevel(item)">{{item.text}}</pre><div><p ng-show="!isDisplayTrace && !isDisplayDebug && !isDisplayWarning && !isDisplayError && !isDisplayOther" class="text-center"><i>(select filter)</i></p></pre>
```
jcabrerizo, line 23: Is not the second `<div ng-if="isFilterContent()">` redundant as the condition is part of the -now- `ng-show`?

jcabrerizo: ignore

jcabrerizo (approved):

ui-modules/app-inspector/app/components/stream/stream.template.html
algairim: No, it is not, this is the fix.

PR #351, opened 2022-07-15
Fix overwrite rule when adding a config key definition

algairim:


ui-modules/blueprint-composer/app/components/util/model/entity.model.js
```diff
@@ -754,7 +754,7 @@ function addConfigKeyDefinition(param, overwrite, skipUpdatesDuringBatch, value)
             delete paramMapped['default'];
         }
         // making sure we assign to empty object to avoid conflicts with config items' properties like `type`
-        allConfig[key] = Object.assign({}, configDef, paramMapped, overwrite ? null : configDef);
+        allConfig[key] = Object.assign({}, configDef, paramMapped, overwrite ? configDef: null);
```
algairim, line 757: This fixes the issue where constrains were not included into a config key if value was pre-populated. Thus, `regex` was ignored in validation, when switching with pre-populated value from the YAML editor to Composer.

iuliana (approved):

PR #363, opened 2022-12-05
Fix/catalog saver defaults

jathanasiou:
Adjustments and some refactoring to allow CatalogSaver to better handle default and initial values for the modal dialog fields.

This ensures the defaults adapt properly to user-defined names given in the Editor, who is the parent of the modal.


ahgittin:
Fix looks good.  Can you squash the commits and force push?  In this case that part of history is unhelpful.

ahgittin:
(I will merge as soon as that is done.)

PR #370, opened 2023-01-17
fix for missing version in 'View in Catalog' link

jathanasiou:
Refactoring to avoid regression from previous changes that caused the "version" URL components of this link to be `undefined`.

### DIFF
--- a/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
+++ b/ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
@@ -208,7 +208,8 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
     };
 
     $scope.getCatalogURL = () => {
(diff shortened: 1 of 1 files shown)


ui-modules/blueprint-composer/app/components/catalog-saver/catalog-saver.directive.js
```diff
@@ -208,12 +208,14 @@ export function CatalogItemModalController($scope, $filter, blueprintService, pa
     };
 
     $scope.getCatalogURL = () => {
+        const urlPartVersion = $scope.config.current.version || $scope.config.version;
```
ahgittin, line 211: is `$scope.config.current` guaranteed to be an object at this point?  if not we risk introducing errors with this change.

i'm usually inclined to be defensive eg

`urlPartVersion = ($scope.config.current || {}).version || $scope.config.version`

and possibly an `if (!urlPartVersion) return '';` (though this is _more_ defensive than it was previously)

ahgittin:
:+1: even better

PR #371, opened 2023-01-19
field type syntax fix

jathanasiou:
Corrected syntax on the `ng-switch` directive so that config fields with non-String type are rendered as intended.


ahgittin:
LGTM

PR #373, opened 2023-03-22
Improves filter layout

nakomis:


nakomis:
Primarily changes from (non-standard) multi-select radio buttons to checkboxes and improves readability and display of highlights. Example before and after screenshots attached

Before:
dropdown-before

After:
dropdown-after

ahgittin:
looks much nicer!  good stuff.  merging.

PR #378, opened 2023-04-13
wrapping long step id text

jathanasiou:
This should fix issues with displaying long IDs for workflow steps without them escaping the panel:

workflow-item-wrapping


duncangrant (approved):
Looks good - thanks @jathanasiou

PR #383, opened 2024-01-25
Updated licenses

duncangrant:
Updated licences ahead of release

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -157,9 +157,15 @@ Runtime dependencies:
     License name: MIT License
     Notice:       Copyright (c) 2011-2016 Twitter, Inc.
   
-  ch.qos.logback:1.2.3: 
+  ch.qos.logback:0.1.5: 
     Project:      Logback Logging Framework
-    Version:      1.2.3
+    Version:      0.1.5
+    Available at: https://github.com/qos-ch/logback-contrib/wiki
+    License name: Eclipse Public License, version 1.0
+  
+  ch.qos.logback:1.2.9: 
+    Project:      Logback Logging Framework
+    Version:      1.2.9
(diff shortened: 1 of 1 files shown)


DEPENDENCIES
```diff
@@ -157,9 +157,15 @@ Runtime dependencies:
     License name: MIT License
     Notice:       Copyright (c) 2011-2016 Twitter, Inc.
   
-  ch.qos.logback:1.2.3: 
+  ch.qos.logback:0.1.5:
```
jcabrerizo, line 160: Again this new reference to an old version

jcabrerizo (commented):
Same old version for `ch.qos.logback:0.1.5` than on brooklyn-dist, could you check where it comes from?
See https://github.com/apache/brooklyn-dist/pull/184#pullrequestreview-1845179321

DEPENDENCIES
duncangrant: Looking at mvn dependency:tree
[INFO] +- org.apache.karaf.features:framework:kar:4.3.8:compile
[INFO] |  +- org.ops4j.pax.logging:pax-logging-logback:jar:2.0.18:compile
[INFO] |  |  +- ch.qos.logback:logback-core:jar:1.2.9:compile
[INFO] |  |  +- ch.qos.logback:logback-classic:jar:1.2.9:compile
[INFO] |  |  +- ch.qos.logback.contrib:logback-jackson:jar:0.1.5:compile
[INFO] |  |  +- ch.qos.logback.contrib:logback-json-core:jar:0.1.5:compile
[INFO] |  |  \- ch.qos.logback.contrib:logback-json-classic:jar:0.1.5:compile

It seems to be a different artifact with it's own versioning

jcabrerizo (approved):

PR #384, opened 2024-03-19
Build the UI using `yarn` instead of `npm`

tbouron:
This allows for faster build as `yarn` downloads the dependencies once, then copies the files locally. So on repetitive builds like on Jenkins (or locally when during the development cycle) the step `yarn install` takes on average `0.34s` instead of aroudn `25s`. Multiple that by the number of UI modules we have, then you get the results below.

For instance, when building on my local machine (Mac M1) only the `ui-modules` folder, i.e. the command: `mvn clean install  -rf :brooklyn-ui-utils`:
- using `npm`, the first build (i.e. no nodule_modules) takes `10:26 mins`. Subsequent builds take `6:29 mins`.
- using `yarn`, the first build (i.e. no nodule_modules) takes `08:34 mins`. Subsequent builds take `5:25 mins`.


ahgittin:
awesome - merged

ahgittin:
for me (also M1) it is now `6:44` for initial and `3:56` for subsequent - wow, what a difference!

PR #385, opened 2024-05-21
Calculates href so ctrl click works

duncangrant:
Currently the **Open in Composer** and **Open in Composer (expanded)** buttons in the catalog don't allow ctrl-click or right click and open in new tab behaviour.  This is because they are links using ng-click and have href set to #.  To fix this we now calculate the links whenever someone toggles the dropdown that displays these buttons.
By calculating the href on the links then ctrl click will now open this in a new tab. 

I haven't found any other instances of this behaviour in Brooklyn so I'm confident that no other changes are required.

I have tested this on a mac but not on MS Windows.


jcabrerizo (approved):
LGTM. Thanks @duncangrant

ahgittin:
looks good, merged with a bit of bringing back some of the previous code, just in case

PR #387, opened 2024-06-28
Use simpler colors for kilt, with option to expand

ahgittin:
This adds a simple color mode to the progress map (kilt diagram), in a settings cog popover which also explains the colors.

I'm thinking the default should maybe be the simple color.  It's saved to local settings so easy enough to enable it once as needed.

Thoughts?


duncangrant (commented):
Looks good - thanks @ahgittin

PR #391, opened 2026-04-21
minor changes in logbook ui

zan-mateusz:
* add a text field for logger name/class prefix filtering (separate CR to handle this in brooklyn-server)
* move the field above and also the date and number of lines filters to a separate line and make them be shown/hidden with the button 'show additonal options' added to the button pallette
Screenshot 2026-04-20 at 10 56 38


jcabrerizo (approved):
LGTM

